PR #1, opened 2026-06-11
Linux: XInitThreads before GTK init — fix intermittent X11 startup crash

nesquena:
## Symptom

The Linux smoke harness failed on **5 of 7 runs of identical app code** (everything after the D17 fix), while the same binary passed twice — including v0.3.1's release run. Two failure shapes, one cause:

- `[xcb] Unknown sequence number while awaiting reply … Assertion '!xcb_xlib_threads_sequence_lost' failed` → SIGABRT
- silent death with exit code 1 (Xlib's default fatal-IO handler calls `exit(1)`)

## Root cause

WebKitGTK's internal threads make direct X11 calls. Xlib is not thread-safe unless `XInitThreads()` is called **before any other Xlib call in the process** — and nothing in our stack (GTK3, tao's gtk backend, this WebKitGTK build) does it for us. Concurrent X traffic from WebKit's threads vs the GTK main loop intermittently corrupts Xlib's reply sequence state at startup. Timing-dependent: CI under Xvfb amplifies it, but a real X11 desktop can hit it.

## Fix

`XInitThreads` via `x11-dl` (dlopen, so Wayland-only systems without libX11 gracefully skip) as the **first statement of `main`**, ahead of GTK/GDK display open. Codified as CLAUDE.md invariant #11.

## Validation

- `cargo fmt` / `clippy -D warnings` / `cargo test` clean (macOS)
- `cargo xwin check` clean (Windows target unaffected — dep and code are Linux-only)
- Linux Smoke dispatched repeatedly on this branch — results posted below before merge. Prior baseline: ~30% pass rate; bar: 5/5 green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,18 @@
 # Changelog
 
+## [v0.3.2] — 2026-06-10
+
+### Fixed
+
+- **Linux: intermittent crash at launch.** WebKitGTK's internal threads talk
+  X11 directly; without Xlib's thread-safe mode they race the GTK main loop
+  and the app could abort during startup (`[xcb] Unknown sequence number
+  while awaiting reply`) or die silently with no window — roughly two out of
+  three launches in the CI smoke harness, timing-dependent on real desktops.
+  The app now calls `XInitThreads` before anything else on Linux. Wayland-only
+  systems without libX11 are unaffected (the call is skipped). Found by the
+  Linux smoke harness flaking on identical builds.
+
 ## [v0.3.1] — 2026-06-10
 
 The first release that arrives **as an in-app update** for v0.3.0 users — and the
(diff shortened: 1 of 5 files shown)


nesquena:
Validation complete: **5/5 Linux Smoke runs green** on this branch ([1](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27318733591), [2](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27318735716), [3](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27318737727), [4](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27318739848), [5](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27318741881)). Baseline on identical pre-fix code was ~30% (5 fail / 2 pass) — P(5/5 by luck) ≈ 0.2%. Merging.

Issue #6, opened 2026-06-12
Show the current desktop version and a 'what's new' / changelog in-app


nesquena-hermes:
## Summary
*"Would be good to see that I'm on the latest desktop version, as well as get some info about what changed somewhere."* (**b3nw**)

## Proposed
- Surface the current app version (e.g. in the menu / About, and the macOS app menu).
- After an auto-update, show a short "what's new" (the release's changelog section). The release pipeline already extracts changelog sections (`scripts/extract_changelog.py`), so the notes are available to surface.

## Files
- `src-tauri/src/menu.rs` (menu / About)
- `src/shell.html` / `src/shared/shell.js` (strip menu)

nesquena-hermes:
Confirmed working on **Windows v0.6.1** by @b3nw (#webui, 2026-06-19):

> #6 yes, works

The in-app version + "What's New"/changelog surface (app menu / ⋯ menu, shipped v0.5.0 PR #29) renders correctly. Closing as done.

Issue #5, opened 2026-06-12
App icon shows a thin white line/fringe around the rounded corners


nesquena-hermes:
## Summary
*"The corner radius on the icon seems to be producing a white line around it."* (**Rod**)

## Notes
The source `src-tauri/icons/icon.png` (512x512) has fully transparent corners (RGBA 0,0,0,0 at all four corners) and a dark navy body -- so the fringe is most likely from one of: the generated `.ico` / Windows tile assets, an opaque matte/halo baked into a downscaled size, or the OS rounding a square icon against a light background rather than the source asset itself. Worth checking the per-size PNGs and the `.ico` for a light matte ring at the rounded edge, and whether any rounded-corner masking is applied with premultiplied alpha.

## Files
- `src-tauri/icons/*` (icon.png, icon.ico, Square*Logo.png set)

nesquena:
Closing as superseded: this was reported against the original navy icon, and **v0.6.7 replaced the entire icon set** with the new cyan caduceus artwork (#61) — freshly generated Win/Mac/Linux assets. I also pixel-checked the new 512×512 source: all four corners are fully transparent (no opaque matte, the mechanism most likely behind the old fringe), and no fringe has been reported in the ~2 weeks since the new icon shipped.

@rod if you still see a light edge around the rounded corners on ≥ v0.6.7 (worth checking the Windows taskbar/tiles specifically, where downscaled .ico frames live), please reopen with a screenshot and we'll chase the generated assets.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #3, opened 2026-06-12
Multi-tab: sidebar shows wrong profile after tab switch; profile switch intermittently drops that profile's default workspace


nesquena-hermes:
## Summary
Two related multi-profile issues in Windows/Linux strip-tabs mode, both reported by **b3nw**:
1. **Switching tabs doesn't update the WebUI sidebar to that tab's profile.** *"When you switch tabs, the profile and session list on the left doesn't change to the new profile the tab is on."*
2. **Switching profiles intermittently fails to apply that profile's default workspace.** *"If I switch profiles 2-3 times, it reliably fails to set the correct workspace at least one of the times -- only on the desktop client."* (intermittent, ~1 in 2-3)

## Likely root cause
Each tab is a separate content webview (`src-tauri/src/strip.rs` `add_tab`), but all tab webviews share one WebView2 user-data dir / WebKitGTK web context -- i.e. a single cookie jar. The WebUI selects the active profile **per-client via the `hermes_profile` cookie** (webui #798). With a shared cookie store, profile is effectively global across tabs: switching profile in one tab flips the cookie for all tabs, so an inactive tab's sidebar goes stale, and switching to another tab shows the latest-cookie profile rather than "that tab's" profile. The intermittent workspace-default miss is consistent with the same shared-state race (profile cookie vs. the per-profile default-workspace read on reload).

## Proposed direction
- Give each tab (or at least each profile) an isolated webview data partition / cookie store so profile selection is per-tab; OR coordinate with WebUI so the active profile can be scoped per connection/tab rather than a single shared cookie.
- For the workspace-default intermittency, confirm whether it's the shared-cookie race above or a separate WebUI timing issue (profile-switch reload racing the default-workspace read) -- may need a companion WebUI fix.

## Files
- `src-tauri/src/strip.rs` (`add_tab`, `select_tab`)
- webview data/cookie configuration (Tauri webview builder / app config)

nesquena-hermes:
Additional confirmation + a third facet, from **b3nw** (Discord, 2026-06-13):

> "If you're in different profiles, not all active chats appear on the sidebar."

> "I'm also still seeing some inconsistent behavior when I change tabs that are in different profiles and/or start new sessions. The most consistent is the workspace often doesn't change, but I've also seen it where the profile chip says one thing but the session seems to be operating in another. I've not been able to pin down the exact action that triggers either of these."

All three symptoms are consistent with the shared-cookie-jar root cause in this issue:

[...]

PR #13, opened 2026-06-12
Fix links not opening in browser (#12) and new-tab white flash (#4)

### LINKED ISSUE #4: White flash when opening a new tab
## Summary
Opening a new tab shows a brief white flash before the page paints. *"White flash when opening tabs hurts the eyes a bit."* (**Rod**)

## Current state
For the first window, anti-flash is handled by building the window hidden and revealing it on first page load (`windows.rs` `open_browser`: `.visible(false)` then `show()` on `PageLoadEvent::Finished`; strip first-window reveal in `strip.rs:213-219`). But in strip mode a *new tab* is a content webview added to an **already-visible** window (`strip.rs add_tab`), so the new webview paints white until first paint -- there is no per-tab pre-paint backstop like the window-level one. The pre-paint background script (`PRE_PAINT`, `bridge.rs:25`) only runs once the webview's document starts, not before the native webview surface shows.

## Proposed fix
- Add the new tab webview hidden and reveal it on its first `PageLoadEvent::Finished` (mirror the window-level anti-flash), keeping the previously-active tab visible until the new one is ready; OR
- Give the new tab webview a pre-painted background matching the cached theme color before reveal.

## Files
- `src-tauri/src/strip.rs` (`add_tab` ~L199-249)
- `src-tauri/src/bridge.rs` (`PRE_PAINT`)

### LINKED ISSUE #12: Clicking a link in the app does nothing (should open in default browser)
## Summary
Clicking a link in chat (e.g. a `https://` URL the agent returned) does nothing in the desktop app — it should open in the system default browser.

Reported by **Deor** in the desktop-apps Discord channel: *"Can clicking a link in the app open the default browser? At the moment it does nothing."*

## Root cause (very likely the same as nesquena/hermes-webui#4040)
The WebUI renders every chat link with `target="_blank"` (markdown anchors + auto-linked bare URLs, `static/ui.js:3562-3563`). In the desktop app the injected `WINDOW_OPEN` script intercepts those clicks, calls `preventDefault()`, and forwards the URL to native via `EMIT('open-external', href)` (`src-tauri/src/bridge.rs:231-239`). The native handler then opens it in the default browser (`bridge.rs:502-508` -> `tauri_plugin_opener::open_url`).

That `EMIT(...)` is a Tauri IPC event that posts to `http://ipc.localhost/plugin:event|emit`. But:
- `src-tauri/tauri.conf.json:13` sets `app.security.csp = null`, so the app does **not** manage the page CSP. When Tauri *does* set a CSP it auto-appends the IPC origin to `connect-src`; with `csp: null` on a **remote** page (the WebUI served from the server) it can't, so the **server's** CSP fully governs the page.
- The WebUI server's `connect-src` does **not** whitelist `ipc.localhost` (see nesquena/hermes-webui#4040). So the IPC POST is blocked, the `open-external` event never reaches Rust, and the click is a no-op.

This is corroborated by a desktop tester's server log showing exactly this URL being blocked/reported:
```
csp_report: ... 'blockedURL': 'http://ipc.localhost/plugin%3Aevent%7Cemit' ...   (= plugin:event|emit)
```
`plugin:event|emit` is precisely the IPC channel `EMIT()` uses.

## Note on plain (non-`target="_blank"`) links
A top-level navigation to an external host is handled separately by `navigation_allowed` (`src-tauri/src/windows.rs:305-333`), which opens externally without using IPC — so that path is unaffected. The broken case is specifically `target="_blank"` links and `window.open()` (which is how WebUI renders chat links), because those go through the IPC `open-external` emit.

## Suggested resolution
- Primary fix is WebUI-side: allow `http://ipc.localhost` in `connect-src` (tracked in nesquena/hermes-webui#4040). Once that lands, re-verify links open here.
- If a desktop-only mitigation is wanted as a fallback, consider whether the wrapper can guarantee the `open-external` path (e.g. handle `target="_blank"` link clicks through the native navigation/opener layer rather than the CSP-governed IPC emit).

## Files
- `src-tauri/src/bridge.rs` (`WINDOW_OPEN` L215-241; `open-external` handler L502-508)
- `src-tauri/src/windows.rs` (`navigation_allowed` L305-333)
- `src-tauri/tauri.conf.json` (`app.security.csp = null`, L12-14)
- Cross-ref: nesquena/hermes-webui#4040 (CSP blocks `ipc.localhost`)

nesquena:
Two low-risk, no-brainer fixes from the open issue tracker.

## #12 — Clicking a link in chat did nothing (should open the browser)

**Symptom:** clicking a `https://` link in chat is a no-op.

**Root cause:** the WebUI renders chat links with `target="_blank"`. The injected `WINDOW_OPEN` script intercepted those clicks and forwarded the URL to native via `EMIT('open-external', href)`. That emit is a Tauri IPC POST to `ipc.localhost`, but the remote page's CSP `connect-src` (governed by the WebUI server, since `app.security.csp` is null) does not whitelist `ipc.localhost` — so the event never reached Rust and the click did nothing. Confirmed by a tester's server log showing `blockedURL: http://ipc.localhost/plugin%3Aevent%7Cemit`.

**Fix:** external links and `window.open` now navigate the top frame (`location.href = …`) instead of emitting. The native `on_navigation` hook (`windows::navigation_allowed`) — which is **not** subject to the page CSP — opens external hosts in the system browser and cancels the navigation so the page stays put. This is the exact path plain (non-`target="_blank"`) external links already used, so it's proven on all three platforms. No server-side change required (the WebUI-side CSP allowance in hermes-webui#4040 is no longer a prerequisite for links to work here).

`bridge.rs` `WINDOW_OPEN`. The now-unused `open-external` Rust handler is left in place as a harmless defensive fallback.

## #4 — White flash when opening a new tab (Windows/Linux)

**Symptom:** opening a new tab shows a brief white flash before the page paints ("hurts the eyes a bit").

**Root cause:** a new tab is a child webview added to an **already-visible** window, so it paints white until first paint. The window-level anti-flash (build hidden → reveal on `PageLoadEvent::Finished`) can't cover a child webview of a live window.

**Fix:** new tab webviews get an opaque native background in the cached theme color (`WebviewBuilder::background_color`), so they come up theme-colored instead of white. On Windows the alpha is ignored and the color renders opaque, which is what we want. macOS uses native tabs (not the strip), so it's unaffected.

`strip.rs` `add_tab`.

## Testing
- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo test` all green locally.
- #12 reasoned against the existing, proven `navigation_allowed` external-open path; #4 is an additive native background hint with no behavior change.

## Not included (deliberately deferred — not no-brainers)
- **#3** multi-tab profile/cookie isolation — needs per-tab webview data partitioning and possibly WebUI coordination.
- **#5** icon white fringe — uncertain root cause; needs icon-asset regeneration + visual verification.
- **#6** in-app version / what's new — the version is already shown (macOS app menu and strip ⋯ menu); only the post-update "what's new" remains, which is larger.

Closes #12
Closes #4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## [Unreleased]
+## [v0.3.4]
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,27 @@
 # Changelog
 
+## [Unreleased]
+
+### Fixed
+
+- **Clicking a link in chat did nothing instead of opening the browser.**
+  The WebUI renders chat links with `target="_blank"`, and our injected script
+  forwarded those clicks to native by emitting an `open-external` IPC event.
+  That emit posts to `ipc.localhost`, which the remote page's own CSP
+  `connect-src` blocks (the WebUI server governs the page's CSP, not the shell),
+  so the event never reached Rust and the click was a no-op. External links
+  (and `window.open`) now navigate the top frame instead: the native
+  `on_navigation` hook — which is not subject to the page CSP — opens external
+  hosts in the system browser and cancels the navigation so the page stays put,
+  exactly as plain links already did. No server-side change required.
+  (#12, reported by Deor; cross-ref hermes-webui#4040.)
+- **White flash when opening a new tab (Windows/Linux).** A new tab is a child
+  webview added to an already-visible window, so it painted white until its
+  first paint — the window-level anti-flash (build hidden, reveal on load)
+  couldn't cover it. New tab webviews now get an opaque native background in the
+  cached theme color, so they come up theme-colored instead of white. (#4,
+  reported by Rod.)
+
 ## [v0.3.3] — 2026-06-10
 
 ### Fixed
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,7 +26,9 @@ state; the shell owns the window, the ssh process, and a small prefs JSON.
 ## Repo structure
 
 ```
-src/                      # shell pages (vanilla): splash.html, error.html, prefs.html
+src/                      # shell pages (vanilla): splash.html, error.html, prefs.html,
+                          #   index.html, shell.html (Win/Linux tab-strip shell) + shared/
+                          #   (shell.css, shell.js)
 src-tauri/src/
   main.rs                 # tauri Builder, plugins, commands, window/menu events
   conn.rs                 # connection orchestrator (splash → connect → windows | error)
@@ -42,6 +44,14 @@ src-tauri/src/
   theme.rs                # CSS color parse / luminance / hex (pure fns, unit-tested)
   paste.rs                # clipboard image → PNG → base64 → 3-strategy injection
   menu.rs                 # macOS menu bar (Win/Linux use the injected shortcut forwarder)
+  state.rs                # shared AppState: TunnelStatus, per-window TabEntry/WindowTabs,
+                          #   ssh Child handles, atomics (mirrors Swift TunnelManager enum)
+  strip.rs                # Win/Linux tab strip: 38px shell.html webview + one live
+                          #   content webview per tab; hidden tabs stay alive (SSE/draft
+                          #   survival). HERMES_FORCE_STRIP=1 exercises it on macOS
+  updater.rs              # auto-update (Sparkle parity): minisign-verified vs pubkey in
+                          #   tauri.conf.json, latest.json manifest; passive + interactive
+                          #   check. Win NSIS/MSI, mac .app, Linux AppImage (.deb degrades)
 src-tauri/capabilities/   # IPC scoping: full for shell pages; remote content may ONLY
                           #   emit events (content.json)
 ```
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -222,20 +222,31 @@ const NOTIFY_WATCHER: &str = r##"
 /// (the WebUI server, not the shell, governs CSP because `app.security.csp` is
 /// null) — so the emit silently fails and the click is a no-op (issue #12,
 /// hermes-webui#4040). `on_navigation` is a native wry hook, not subject to CSP.
+///
+/// Only http(s) URLs are routed through the top-frame navigation:
+/// `navigation_allowed` deliberately lets non-web schemes through (the engine
+/// needs them internally), so navigating to `about:blank`/`blob:`/`javascript:`
+/// here would REPLACE the app's top frame (a common `window.open` pattern in
+/// libraries). Non-http(s) window.open calls are dropped (the old EMIT path's
+/// effective behavior); non-http(s) `_blank` anchors keep their default action
+/// so the download bridge / engine still handle them.
 const WINDOW_OPEN: &str = r##"
   (function () {
     window.open = function (u) {
       if (!u) return null;
-      location.href = String(u);
+      try {
+        var x = new URL(String(u), location.href);
+        if (x.protocol === 'http:' || x.protocol === 'https:') location.href = x.href;
+      } catch (e) {}
       return null;
     };
     document.addEventListener('click', function (e) {
       var a = e.target && e.target.closest ? e.target.closest('a[target="_blank"]') : null;
-      if (a && a.href) {
-        e.preventDefault();
-        e.stopPropagation();
-        location.href = a.href;
-      }
+      if (!a || !a.href) return;
+      if (a.protocol !== 'http:' && a.protocol !== 'https:') return;
+      e.preventDefault();
+      e.stopPropagation();
+      location.href = a.href;
     }, true);
   })();
 "##;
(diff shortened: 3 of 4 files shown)


nesquena:
# Deep review (end-to-end trace) — with fixes applied directly

**Verdict: ready to merge.** Both root-cause analyses are correct — verified against the tester evidence, the hermes-webui source, and the locked tauri/wry sources. Two review findings were worth changing before merge; rather than bounce the PR, they're applied in **2d0ea6e** (details below). CI on the final head: Test ✅, Linux Smoke ✅ ([run 27448485891](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27448485891)).

## What was verified

### #12 — external links
The root cause holds end-to-end: `EMIT` → `window.__TAURI__.event.emit` → fetch to `http://ipc.localhost/plugin:event|emit`, blocked by the server-governed CSP (`app.security.csp` is null so the shell never augments `connect-src`; the tester's `blockedURL` log is direct proof). The replacement routes through a top-frame navigation, and `navigation_allowed` is a native wry policy hook — CSP can't touch it. Every URL class traced through old vs new:

| Input | Old | New |
|---|---|---|
| External http(s) `_blank` / `window.open` | EMIT → **CSP-blocked → no-op** (the bug) | `location.href` → native cancel + system browser ✅ |
| Internal/relative `_blank` / `window.open` | `location.href` (top-frame navigate) | **identical branch — no regression** |
| `blob:` / `about:` / `javascript:` via `window.open` | EMIT → dropped | guarded (see fix 1) |

Checked against hermes-webui @ c77c8312: every `_blank` render site (`ui.js` 848/3557/3763/3952/5598/5776) and both `window.open` call sites (`ui.js:639` dashboard, `workspace.js:819` open-in-browser) are http(s)/relative — so current webui has no behavior change for internal links, and chat links (the bug) now work. `DOWNLOAD_BRIDGE` interplay is safe: export anchors carry `download`/`blob:` without `_blank`, and both capture listeners sit on the same node so `stopPropagation()` can't suppress the download path.

### #4 — new-tab flash
`pre_paint_color` returns f64 0..1 → the `to8` mapping is correct (default dark → `Color(26,26,26,255)`). Verified at the locked versions: `WebviewBuilder::background_color` exists (tauri 2.11.2), and wry 0.55.1 applies it **at creation time** on WebKitGTK — no post-creation geometry, so the known GTK child-webview crash constraint is respected. WebView2 clamps non-zero alpha to opaque. Unclaimed bonus: the *first* tab also flows through `add_tab`, so the 4s show-fallback path (page never finishes) now reveals theme color instead of white too. The Linux smoke harness exercises exactly that codepath (first tab = `add_tab` with `background_color` on a GTK child webview): dispatched on this branch twice, green both times ([pre-fix head](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27444545747), [final head](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27448485891)).

## Fixes applied in 2d0ea6e

1. **http(s)-only guard in `WINDOW_OPEN`.** `navigation_allowed` deliberately admits non-web schemes (the engine needs `about:`/`blob:`/`data:` internally), so the unguarded `location.href` would let a single future `window.open('about:blank')` — a very common library print/preview idiom — **replace the app's top frame**. Now only http(s) routes through the navigation; non-http(s) `window.open` is dropped (the old EMIT path's effective behavior), and non-http(s) `_blank` anchors keep their default action so the download bridge still sees `blob:`/`download` anchors (strictly safer than both old and new code on Windows, which has no download bridge).
2. **CHANGELOG section retitled `[Unreleased]` → `[v0.3.4]`.** `bump_version.sh` seeds its section immediately under `# Changelog` and only skips when the exact version header exists — with an `[Unreleased]` header, the next release would ship an **empty** notes section with these entries stranded above it. (If this ships under a different version, retitle at bump time.)
3. Comment accuracy nit in `strip.rs`: wry *clamps* non-zero alpha to opaque on Windows rather than ignoring it.

## Follow-ups (not blocking; not filed as issues — say the word and I'll open them)

- **Windows terminal interaction to verify:** webui's `terminal.js:648` registers `beforeunload` that closes the terminal SSE and beacons `api/terminal/close`. On Chromium/WebView2, renderer-initiated navigations fire `beforeunload` *before* the host's cancel — so clicking an external chat link with the terminal panel open likely tears down the terminal session even though the page stays. Bounded (terminal only — chat SSE has no `beforeunload` hook, so in-flight chats are safe) and best fixed webui-side (`beforeunload` → `pagehide`, which only fires on committed unloads). Needs one repro on a real Windows box.
- **Residual CSP scope:** this PR fixes links, but every other injected feature still rides EMIT and stays CSP-blocked on current servers — theme reports, title watcher, notifications, downloads, and the Win/Linux in-page shortcut forwarder (Ctrl+T/N/W). hermes-webui#4040 is still wanted for those.
- **Parity-plus idea (pre-existing behavior, unchanged here):** same-origin `window.open`/`_blank` (open-in-browser, HTML file badges, dashboard) replaces the app view today; in strip mode these could open a real new tab.

Issue #7, opened 2026-06-12
Allow renaming tabs


nesquena-hermes:
## Summary
*"Tabs are great, being able to rename them would be very useful."* (**Deor**; +1 from **b3nw**)

Currently tab titles are derived from the page `document.title` (`strip.rs set_tab_title` -> `windows::display_title`). Allow a user-set custom label that overrides the derived title (persist per tab/window; survive title-watcher updates).

## Files
- `src-tauri/src/strip.rs` (`set_tab_title`, `TabEntry.title`, `emit_tabs`)
- `src/shared/shell.js` / `shell.html` (tab UI: e.g. double-click to rename)

nesquena:
Implemented in **v0.5.0**: double-click a tab title to rename it inline (`startRename` in `shell.html` → `tab_rename` → `strip::rename_tab`); the custom name overrides the page title and persists/restores across restart (`custom_title` in `SessionTab`). The "edit box vanishes while typing" regression (#38) was fixed in v0.6.2 (shipped in v0.6.3). Verified in code + round-trip test.

PR #16, opened 2026-06-14
Fix Windows/Linux tabs stuck on "New Tab": source titles natively (#15)

### LINKED ISSUE #15: Windows/Linux tabs stuck on "New Tab" — page→native title bridge (EMIT) doesn't reach native from remote webviews
## Symptom

Windows/Linux desktop tabs don't reflect the active session. Two reported shapes (both on the custom tab strip):

- **"New Tab" never updates** (Deor): every tab shows the literal placeholder `New Tab`.
- **Title resets to the host fallback** (b3nw): switching sessions briefly shows the right title, then it "instantly re-sets back to default" — every tab + window title collapse to `Hermes WebUI ● <host>` (see screenshot in #webui). That string is `windows::display_title()`'s **empty-strip fallback**, which only fires when the page-reported title strips to empty.

## This is desktop-shell-only, not a WebUI regression

The custom HTML tab strip is **Windows/Linux only** — `strip::enabled() = cfg!(not(target_os = "macos"))`. macOS uses native WKWebView tabs and is unaffected (confirmed working on Mac latest). The strip is new (landed v0.2.0, 2026-06-10).

I drove the **current** WebUI (`v0.51.414`) live and instrumented every `document.title` write across welcome → session-select → panel-switch → back. It always emits a well-formed `"<session-title> — Hermes"` (and `"Untitled — Hermes"` when a session has no title). So the WebUI side is healthy; the recent WebUI titlebar commits (#4039 panel-switch sync, #4121 `● ` pending-prompt badge, e5640a42 ownership fix) do **not** produce an empty title. The failure is in the strip's title pipeline consuming/displaying it.

## Pipeline (for whoever picks this up)

`bridge.rs TITLE_WATCHER` (MutationObserver on `<title>` + 2s `setInterval`) → `EMIT('title', t)` → `bridge install` handler → `strip::set_tab_title` → `windows::display_title(raw, mode, target, healthy)` → `tab.title` + `emit_tabs` + `refresh_window_title`.

Two distinct failure modes map cleanly onto two pipeline breaks:

1. **`EMIT('title')` never reaches native** → tab stays at the `"New Tab".into()` placeholder (strip.rs ~L252). `EMIT` is a silent `try{…}catch{}` that no-ops unless `window.__TAURI__.event` is present and the `bridge` event is granted for the tab's **remote** origin (`capabilities/content.json`). Worth verifying the capability covers tunneled/non-localhost origins (the affected users connect to remote/SSH targets) and that `TITLE_WATCHER` is actually injected into tab webviews (it is wired in `add_tab` via `init_script`, but confirm timing vs. page load).

2. **Page reports a title that `display_title` strips to empty** → falls through to `format!("Hermes WebUI  {dot} {host}")`. The strip regex `\s+[—\-|·]\s+Hermes(\s+Agent)?\s*$` only removes a trailing ` — Hermes`/` — Hermes Agent`. Two coupling fragilities here:
   - The WebUI appends ` — <assistantDisplayName()>`, which is **not** always `"Hermes"`: it's the user's configured bot name (configurable since the "make bot name configurable" change) or, on a **non-default profile**, the capitalized profile name. When that suffix isn't literally `Hermes`, the regex won't strip it (full string shown, not empty) — so empty-strip implies the page genuinely reported an empty/separator-only title for that tab (e.g., a tab mid-load, or a state where `document.title` is briefly just the separator). The "instant reset" timing suggests the 2s poller or a re-report is catching a transient empty state and the strip isn't guarding against regressing a previously-good title to the fallback.

## Suggested fixes (desktop side)

- **Don't downgrade a known-good tab title to the host fallback on a transient empty report.** In `set_tab_title`, if `display_title(raw,…)` would yield the fallback but the tab already has a real (non-fallback) title, keep the previous one (or ignore empty `raw`).
- **Decouple from the hardcoded ` — Hermes` suffix.** The regex breaks for custom bot names and non-default profiles. Prefer stripping the *separator + trailing segment* generically, or have the WebUI expose a clean title via a dedicated bridge channel rather than reverse-engineering `document.title`.
- **Verify the `bridge`/`title` IPC capability is granted for remote/tunneled tab origins** so `EMIT('title')` isn't silently dropped (covers the "New Tab forever" mode).

## Ask for reporters

Please confirm **OS + desktop app version** (`gh release list` latest is desktop **v0.3.4**). Strong likelihood at least one report is on a pre-v0.3.4 / pre-strip-fix build — version skew would explain why Mac-latest is clean.

Cross-ref: WebUI side verified healthy at nesquena/hermes-webui `v0.51.414`. Related desktop attention-surfacing ask: #14.

nesquena:
Fully resolves #15 — Windows/Linux tab strip titles never update (stuck on the literal "New Tab"), and an earlier-build variant where titles "instantly reset" to `Hermes WebUI ● <host>`.

## Root cause

The strip's only title source was an injected script (`TITLE_WATCHER`) that posted `document.title` back to native over the JS event IPC:

```
TITLE_WATCHER → EMIT('title', t) → window.__TAURI__.event.emit('bridge', …) → bridge handler → set_tab_title
```

`EMIT` is a silent `try{…}catch{}` that no-ops unless `window.__TAURI__.event` is present and the event IPC is accepted for the tab's **remote** origin. In remote-origin content webviews that path isn't reliable, so the title report silently dropped and the tab kept its `"New Tab".into()` seed (issue failure mode #1). This is the same class of failure the #12 link fix hit — and it fixed it by **not** depending on the remote webview's IPC. The title watcher was never migrated.

macOS was unaffected only because the strip is Windows/Linux-only (`strip::enabled() = cfg!(not(target_os = "macos"))`); macOS uses native WKWebView tabs.

## Fix

**Source the title natively.** Wire wry's native `on_document_title_changed` hook (Tauri 2.11 `WebviewBuilder`/`WebviewWindowBuilder`) on both the strip tab webviews (`strip.rs add_tab`) and the macOS content windows (`windows.rs open_browser`). It fires on the engine's own title-changed signal — implemented on WebView2, WebKitGTK, and WKWebView — so it's independent of page JS, `window.__TAURI__`, and the page CSP. The JS `TITLE_WATCHER` and the now-dead bridge `"title"` handler arm are removed; both paths converge on the new `windows::apply_reported_title`.

**Three more correctness fixes from the issue's analysis:**

1. **Transient-empty guard** (failure mode #2). `apply_reported_title` ignores a report that cleans to empty/separator-only, so a known-good title — or the initial "New Tab" seed — is never downgraded to the `Hermes WebUI ● <host>` fallback while a tab is mid-load.
2. **Generic suffix stripping.** `display_title` stripped a hardcoded ` — Hermes`/` — Hermes Agent`. But the WebUI appends the **configured bot name** or, on a non-default profile, the **profile name**. Now the trailing separator-delimited segment is stripped generically (only the last one, so an em-dash inside the session title itself survives).
3. **Capability covered plain HTTP remote origins.** `content.json` listed `https://*` but no `http://*`, so for a Direct-mode connection to a non-localhost HTTP server (e.g. `http://192.168.x.x:8787`) the capability didn't apply and *every* bridge emit was dropped — including the theme sync and the "response is ready" notification (the issue's corroborating prediction that notify is also dead). Now `http://*`/`http://*:*` are covered. Still event-emit-only — no command access (invariant #7 preserved).

## Why native is the right call (vs. "repair `window.__TAURI__` injection")

The title only needs the page's title string, which the engine already tracks — no IPC required. The native hook can't be defeated by CSP, capability-scope gaps, or missing IPC globals, so it resolves failure mode #1 regardless of which of those is the culprit on a given user's setup. The capability fix (#3) separately restores the *other* bridge emits that genuinely need the IPC.

## Testing

- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo test` all green.
- Added unit tests: generic suffix stripping (custom bot name, profile name, internal em-dash preserved) and the transient-empty cases (`""`, whitespace, `" — Hermes"`, and that a bare `"Hermes"` is treated as meaningful).
- Native title hook confirmed implemented on all three desktop engines in the locked wry 0.55.1 (`webview2`/`webkitgtk`/`wkwebview`).

## Note for reviewers / verification

The strip is Windows/Linux-only; `HERMES_FORCE_STRIP=1` exercises it on macOS for a local check. A real Windows/Linux smoke (open a couple of sessions, switch tabs, confirm titles track the active session and don't reset) is the high-value manual check before publish.

Closes #15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,36 @@
 # Changelog
 
+## [v0.3.5]
+
+### Fixed
+
+- **Windows/Linux: tabs were stuck on "New Tab" and never showed the active
+  session (and titles could "reset" to `Hermes WebUI ● <host>`).** The custom
+  tab strip derived each tab's title from an injected script that posted the
+  page's `document.title` back over the JS event IPC (`window.__TAURI__`). That
+  IPC isn't reliably available in remote-origin content webviews, so the title
+  report silently no-op'd and the tab kept its placeholder. Titles are now read
+  from the webview engine directly via a native title-changed hook
+  (WebView2/WebKitGTK/WKWebView), independent of page JS, the IPC global, and
+  the page's CSP — the same "don't depend on the remote webview's IPC" approach
+  the link-opening fix used. (#15, reported by Deor and b3nw.)
+- **A transient blank title no longer wipes a good tab title.** While a tab is
+  mid-load (or briefly reports a separator-only title), the strip used to fall
+  back to the `Hermes WebUI ● <host>` placeholder. A known-good title (or the
+  initial "New Tab" seed) is now kept until a real title arrives.
+- **Tab titles now strip the trailing name suffix for custom bot names and
+  non-default profiles.** The WebUI appends a " — name" suffix where the name is
+  the configured bot name or the profile name — not always "Hermes". The suffix
+  is now removed generically (only the last separator-delimited segment, so an
+  em-dash inside the session title itself is preserved) instead of matching the
+  literal "Hermes".
+- **Direct-mode connections to a non-localhost HTTP server now get the bridge
+  again.** The content-webview capability listed `https://*` but no plain
+  `http://*`, so for a Direct connection to e.g. `http://192.168.x.x:8787`
+  every bridge emit (theme sync, the "response is ready" notification) was
+  silently dropped. Plain-HTTP remote origins are now covered (still
+  event-emit-only — no command access).
+
 ## [v0.3.4] — 2026-06-12
 
 ### Fixed
--- a/src-tauri/capabilities/content.json
+++ b/src-tauri/capabilities/content.json
@@ -10,6 +10,8 @@
       "http://localhost:*",
       "http://127.0.0.1",
       "http://127.0.0.1:*",
+      "http://*",
+      "http://*:*",
       "https://*",
       "https://*:*"
     ]
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -251,25 +251,6 @@ const WINDOW_OPEN: &str = r##"
   })();
 "##;
 
-/// Title watcher — feeds the native tab/window title pipeline.
-const TITLE_WATCHER: &str = r##"
-  (function () {
-    let last = null;
-    const report = function () {
-      const t = document.title || '';
-      if (t !== last) { last = t; EMIT('title', t); }
-    };
-    const start = function () {
-      report();
-      const el = document.querySelector('title');
-      if (el) new MutationObserver(report).observe(el, { childList: true, characterData: true, subtree: true });
-      setInterval(report, 2000);
-    };
-    if (document.readyState === 'loading') document.addEventListener('DOMContentLoaded', start);
-    else start();
-  })();
-"##;
-
 /// SSH footer — 28px status bar pinned to the bottom, tinted with the exact
 /// page background (Swift parity: the footer is the only chrome painted with
 /// the page RGB). Injected only in ssh-mode windows. The native side drives
@@ -445,7 +426,6 @@ pub fn init_script(label: &str, pre_paint_hex: &str, is_ssh: bool) -> String {
     parts.push(THEME_BRIDGE);
     parts.push(NOTIFY_WATCHER);
     parts.push(WINDOW_OPEN);
-    parts.push(TITLE_WATCHER);
     parts.push(FIND_BAR);
     if cfg!(any(target_os = "macos", target_os = "linux")) {
         parts.push(DOWNLOAD_BRIDGE);
@@ -481,16 +461,10 @@ pub fn install(app: &AppHandle) {
         let label = payload["label"].as_str().unwrap_or("").to_string();
         let kind = payload["kind"].as_str().unwrap_or("");
         match kind {
-            "title" => {
-                let raw = payload["value"].as_str().unwrap_or("").to_string();
-                if label.starts_with("tab-") {
-                    crate::strip::set_tab_title(&handle, &label, &raw);
-                } else {
-                    let state = handle.state::<AppState>();
-                    state.raw_titles.lock().unwrap().insert(label.clone(), raw);
-                    windows::refresh_title(&handle, &label);
-                }
-            }
+            // NOTE: titles no longer arrive via this bridge — they're sourced
+            // natively through wry's on_document_title_changed hook (see
+            // windows::apply_reported_title), which works even when the
+            // remote-webview IPC that powers EMIT() is unavailable (issue #15).
             "theme" => {
                 if let Some(css) = payload["value"].as_str() {
                     handle_theme_report(&handle, css);
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -210,6 +210,12 @@ pub fn add_tab(app: &AppHandle, window_label: &str) {
         .on_navigation(move |url| {
             windows::navigation_allowed(&nav_app, url, allowed_host.as_deref())
         })
+        // Native title source — replaces the JS EMIT('title') watcher, which
+        // silently no-ops in remote-origin tab webviews (issue #15). The
+        // webview's label IS the tab label.
+        .on_document_title_changed(|wv, title| {
+            windows::apply_reported_title(wv.app_handle(), wv.label(), &title);
+        })
         .on_page_load(move |webview, payload| {
             if !matches!(payload.event(), tauri::webview::PageLoadEvent::Finished) {
                 return;
--- a/src-tauri/src/windows.rs
+++ b/src-tauri/src/windows.rs
@@ -166,6 +166,11 @@ pub fn open_browser(app: &AppHandle, p: &prefs::Prefs, as_tab: bool) -> Option<W
         .visible(false)
         .initialization_script(&init)
         .on_navigation(move |url| navigation_allowed(&nav_app, url, allowed_host.as_deref()))
+        // Native title source (see apply_reported_title) — replaces the JS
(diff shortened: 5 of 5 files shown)


nesquena:
# Deep review (end-to-end trace + native verification)

**Verdict: approve — merge as-is. No blocking issues; no changes applied.**

This one implemented #15's analysis cleanly and completely, so unlike #13 there was nothing in the correctness or release-breaking class to fix — and I'd rather say that plainly than churn a solid PR. Everything below is verification receipts + a few optional notes.

## Verified end-to-end

**The load-bearing claim — the native hook exists and fires on every desktop engine** — checked at the locked sources, not assumed:
- `on_document_title_changed` is present on **both** builders in tauri 2.11.2: `WebviewBuilder` (`webview/mod.rs:594`, cb `Fn(Webview, String)`) and `WebviewWindowBuilder` (`webview/webview_window.rs:326`, cb `Fn(WebviewWindow, String)`) — so both the strip-tab path and the macOS content-window path are valid.
- wry 0.55.1 implements `document_title_changed_handler` on `webview2`, `webkitgtk`, and `wkwebview` — all three desktop engines, so it's genuinely engine-sourced and independent of `window.__TAURI__`, the page CSP, and capability scope. That's the correct fix for #15 failure mode #1, and it's robust no matter which of those was the actual culprit on a given user's box.

**End-to-end on real WebKitGTK (not just unit tests):** dispatched the Linux smoke on this branch ([run 27515304055](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27515304055), green). The fake server serves `<title>Smoke chat — Hermes</title>`, and the screenshot artifact shows the tab rendered as **"Smoke chat"** — the native hook fired, `apply_reported_title` → `clean_title` stripped the suffix, and the strip painted the right title with zero JS IPC. Pre-fix, that exact scenario is the "New Tab forever" bug. The log is panic-free and also shows the *theme* IPC emit still working on the localhost origin — consistent with the diagnosis that the breakage was specifically remote/non-localhost IPC.

**Trace findings (all correct):**
- **Label dispatch** — the webview label *is* the tab/window label, so `apply_reported_title` routes `tab-*` → `strip::set_tab_title` and `main-*` → `refresh_title` correctly on each platform.
- **`clean_title` regex** — traced `"A — B — Hermes"` (→ `"A — B"`, internal em-dash preserved), `" — Hermes"` (→ empty), `"Hermes"` (→ kept), custom bot name / profile name (→ stripped). The `\s*` (not `\s+`) lead and the separator-only collapse are both correct, and the existing `strips_hermes_suffix_and_truncates` test still passes.
- **Transient-empty guard** — the lock guard in `apply_reported_title` is a statement-temporary, dropped before `refresh_title`, so no re-entrant lock; `set_tab_title`'s critical sections are likewise non-nested. No deadlock.
- **Threading / invariant #12** — `on_document_title_changed` fires from inside the wry callout (same context as the existing `on_navigation`/`on_page_load` hooks this code already calls `show()`/`set_zoom()`/`refresh_macos_chrome` from). `set_title` doesn't force a cross-window synchronous redraw the way `addTabbedWindow` does, so it's not in invariant #12's hazard class — consistent with proven usage.
- **Capability `http://*` / `http://*:*`** — necessary and correct: a Direct connection to e.g. `http://192.168.x.x:8787` matched no prior URL pattern (only `https` had a wildcard), so every bridge emit dropped. Still `core:event:default` only — invariant #7 (event-emit, no commands) preserved. Effective exposure is unchanged because `navigation_allowed` independently gates what can load (localhost/configured host; everything else opens externally + cancels); this just brings `http` to parity with the already-present `https://*`.
- **Clean removal** — no dangling refs to `TITLE_WATCHER` or the bridge `"title"` arm; `raw_titles` is still fed (by `apply_reported_title` on macOS, `set_tab_title` on the strip).

**Gates:** `cargo fmt`/`clippy --all-targets -D warnings`/`cargo test` (8 pass, +2 new) all green locally; `cargo xwin check` (Windows target) clean — the cfg-gated `strip.rs`/`windows.rs` paths compile.

## Optional notes (none blocking)

1. **Regex recompiled per call, now ×2 per title change** — `clean_title` does `Regex::new(...).unwrap()` and is called once in the guard and again inside `display_title`. The WebUI toggles a `● ` pending badge in `document.title` during streaming, so title-changed can fire frequently. Pre-existing pattern (the old `display_title` did the same), but since this PR rewrites the function, hoisting to a `LazyLock<Regex>` is a clean win. Say the word and I'll push it.
2. **Bot names containing a separator char won't suffix-strip.** If the configured bot name is e.g. `Hermes-Bot` or `a|b`, the trailing-segment regex can't anchor (the inner separator breaks `[^sep]+`), so the full `"Session — Hermes-Bot"` would show. Graceful (shows the real title, no crash) and still strictly better than the old literal-`Hermes` regex; the only perfect fix is the WebUI exposing a clean title over a dedicated channel (noted as the alternative in #15).
3. **Transient-empty guard can briefly leave a stale title.** Navigating from a titled session to a screen whose title cleans to empty keeps the previous title until a real one arrives. This is the deliberate #15-failure-mode-2 tradeoff (better than flashing the host fallback), and the WebUI's welcome reports `"Untitled — Hermes"` → `"Untitled"` (non-empty) anyway, so the window is narrow.
4. The separator set is duplicated between the regex `[—–\-|·]` and the collapse check `"—–-|·"` — keep them in sync if either changes (a shared const would prevent drift).

## Recommendation

Merge as-is. The fix is correct, engine-verified on all three platforms, proven end-to-end on real WebKitGTK, and preserves the IPC-scope invariant. The PR's own call-out stands as the one remaining gate: a real Windows/Linux session-switch smoke (open a couple of sessions, switch tabs, confirm titles track the active session and don't reset) before publish — the automated smoke proves the path runs and renders a stripped title, but not multi-session switching. This is a natural v0.3.5.

PR #17, opened 2026-06-15
Badge Windows/Linux tabs with a pending approval/clarify popup (#14)

### LINKED ISSUE #14: Highlight/badge a tab that has a pending approval or clarify (question) popup
## Summary

In Windows/Linux strip-tabs mode, when a tool/command **approval** popup (or a **clarify / question** popup) fires in a tab that is **not currently active**, the user gets no durable signal that that tab needs attention. Reported by **b3nw** (Discord, 2026-06-13):

> "when a permission popup occurs in a tab that is not active, you see it briefly appear and vanish, I then have to go through and check each tab to find it. Could any tab with a waiting permission popup or question popup be highlighted in some way?"

So the user has to click through every tab to find which session is blocked waiting on input.

## What's happening

Each tab is its own content webview (`src-tauri/src/strip.rs`). When a background tab raises an approval/clarify card, it's rendered inside that hidden webview and persists there — but the active tab only catches a momentary flash as the shell composites, then nothing. There's no per-tab "needs attention" affordance in the tab strip.

This is **distinct** from:
- webui #4041 (approval card *flicker/hang* in a single view — already fixed/closed), and
- rust #8 (per-tab *profile* indicator — a different signal: which profile, not "this tab is blocked").

## Proposed direction

Add a per-tab **attention badge/highlight** in the tab strip for any tab whose session has a pending approval or clarify/question prompt (e.g. a dot, glow, or count on the tab; optionally bring it forward / flash the taskbar entry).

Requires the shell to learn that a (possibly background) tab has a pending prompt. Two ways:
- **WebUI companion signal** (cleanest): the page already knows its pending-approval / pending-clarify state. Expose it to the shell via the existing bridge (a `postMessage`/IPC event, or fold it into the `document.title` the wrapper already mirrors) so the strip can badge the originating tab. May warrant a small companion issue on `nesquena/hermes-webui` to emit that signal.
- **Shell-side heuristic**: have the wrapper observe the approval/clarify DOM in each tab's webview (more brittle).

## Files
- `src-tauri/src/strip.rs` (`TabEntry`, `emit_tabs` payload — add an `attention`/`pending` flag)
- `src/shared/shell.js` / `shell.html` / tab CSS (render the badge/highlight)
- bridge plumbing for the per-tab pending-prompt signal (and, likely, a small WebUI-side emit)

## Reported by
**b3nw** — Discord #report-bugs, 2026-06-13.

nesquena:
Implements #14 — badge Windows/Linux strip tabs that have a pending approval or clarify/question popup, so a blocked background session is findable at a glance instead of clicking through every tab.

> "when a permission popup occurs in a tab that is not active, you see it briefly appear and vanish, I then have to go through and check each tab to find it." — b3nw

## How it works (no extra WebUI work needed)

The signal already exists in the WebUI: `ui.js` `syncTopbar` prepends a `● ` marker to `document.title` when the view's active session has a pending approval/clarify (`activeSessionHasPendingPromptAttention()`, #4121). Each tab is its own webview reporting its own title, so it's a natural per-tab signal.

The v0.3.5 fix (#15) already moved tab titles onto wry's native `on_document_title_changed` hook, which fires for **background** webviews too. So the shell now sees the marker per-tab with zero extra polling — this PR just consumes it.

## Changes

- **`windows::split_attention`** strips the leading `● ` marker and returns the pending flag. `apply_reported_title` threads it through, so the marker no longer leaks into the displayed title text — and the macOS window title stays clean too.
- **`TabEntry`** gains an `attention: bool`, serialized in the existing `tabs-changed` payload.
- **`set_tab_title`** sets the per-tab flag alongside the (marker-free) title.
- **`shell.html`** renders an amber attention dot + a subtle row tint for flagged tabs, and tweaks the tooltip ("… — waiting for you").

The flag tracks the title both ways: it sets when the popup appears (title gains `● `) and clears when it's resolved (title loses it).

## Scope / notes

- Windows/Linux strip only — that's where the issue lives. macOS uses native tabs (no custom strip); it still benefits from the marker being stripped out of the window title.
- Taskbar flash / bring-to-front (mentioned as "optional" in the issue) is intentionally left as a follow-up to keep this focused on the core strip badge.
- Badge color is amber `#ff9f0a` with `rgba()` (not `color-mix`) so it renders identically across WebView2/WebKitGTK regardless of engine version.

## Testing

- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo test` all green (9 pass).
- Added `split_attention` unit test (marker present/absent, and that the suffix-strip still yields a clean title after the marker is removed).
- Verified the marker source in the sibling `hermes-webui` checkout (`static/ui.js:6279`).

**Verification gap for review:** the badge-rendering path is logic-tested but not exercised against a live pending-approval popup (requires staging a real approval in a background tab). A quick Win/Linux check — trigger a tool approval in a non-active tab, confirm that tab lights up and clears on answer — is the high-value manual smoke before publish. The strip can be exercised on macOS via `HERMES_FORCE_STRIP=1`.

Closes #14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## [v0.3.6]
+
+### Added
+
+- **Windows/Linux: tabs now badge when a session is waiting on you.** When a
+  background tab raises a tool-approval or clarify/question popup, you used to
+  see only a momentary flash and had to click through every tab to find the
+  blocked one. Such tabs now show an amber attention dot (and a subtle tint) in
+  the tab strip, so the session waiting for input is obvious at a glance. The
+  signal is the WebUI's existing pending-prompt title marker, read per-tab
+  through the native title hook — no extra polling. (#14, reported by b3nw.)
+
 ## [v0.3.5] — 2026-06-14
 
 ### Fixed
--- a/src-tauri/src/state.rs
+++ b/src-tauri/src/state.rs
@@ -17,6 +17,11 @@ pub enum TunnelStatus {
 pub struct TabEntry {
     pub label: String,
     pub title: String,
+    /// The tab's session has a pending approval/clarify popup waiting on the
+    /// user. Derived from the WebUI's leading "● " title marker (issue #14);
+    /// the strip renders an attention badge so a blocked background tab is
+    /// findable without clicking through every tab.
+    pub attention: bool,
 }
 
 /// Per-window tab state for strip mode.
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -256,6 +256,7 @@ pub fn add_tab(app: &AppHandle, window_label: &str) {
         entry.tabs.push(TabEntry {
             label: tab_label.clone(),
             title: "New Tab".into(),
+            attention: false,
         });
         entry.active = entry.tabs.len() - 1;
     }
@@ -447,8 +448,10 @@ pub fn layout(app: &AppHandle, window_label: &str) {
     }
 }
 
-/// Title reported by a tab's title-watcher script.
-pub fn set_tab_title(app: &AppHandle, tab_label: &str, raw: &str) {
+/// A tab reported a new title (marker-free) and its pending-attention state.
+/// Called from `windows::apply_reported_title`, which sources both from wry's
+/// native title-changed hook and has already stripped the "● " marker.
+pub fn set_tab_title(app: &AppHandle, tab_label: &str, title: &str, attention: bool) {
     let Some(window_label) = window_of_tab(tab_label) else {
         return;
     };
@@ -457,13 +460,14 @@ pub fn set_tab_title(app: &AppHandle, tab_label: &str, raw: &str) {
         .raw_titles
         .lock()
         .unwrap()
-        .insert(tab_label.to_string(), raw.to_string());
+        .insert(tab_label.to_string(), title.to_string());
     {
         let mut strip = state.strip.lock().unwrap();
         if let Some(entry) = strip.get_mut(&window_label) {
             if let Some(tab) = entry.tabs.iter_mut().find(|t| t.label == tab_label) {
                 let p = prefs::load(app);
-                tab.title = windows::display_title(raw, &p.connection_mode, &p.target_url, true);
+                tab.title = windows::display_title(title, &p.connection_mode, &p.target_url, true);
+                tab.attention = attention;
             }
         }
     }
--- a/src-tauri/src/windows.rs
+++ b/src-tauri/src/windows.rs
@@ -7,6 +7,24 @@ use std::sync::atomic::Ordering;
 use std::sync::LazyLock;
 use tauri::{AppHandle, Emitter, Manager, WebviewUrl, WebviewWindow, WebviewWindowBuilder};
 
+/// The WebUI prepends this glyph (U+25CF + space) to `document.title` when the
+/// view's active session has a pending approval/clarify popup waiting on the
+/// user (hermes-webui `ui.js` `syncTopbar`, #4121). Each tab is its own webview
+/// reporting its own title, so this is a per-tab "needs attention" signal — the
+/// input for the strip's attention badge (issue #14).
+const ATTENTION_MARKER: char = '●';
+
+/// Split a reported `document.title` into (pending-attention, marker-free
+/// title). The marker is stripped so it never leaks into the displayed title
+/// text (the strip renders a proper badge instead, and the macOS window title
+/// stays clean).
+pub fn split_attention(raw: &str) -> (bool, &str) {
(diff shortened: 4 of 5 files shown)


nesquena:
# Deep review (end-to-end trace + codepoint verification + visual render)

**Verdict: approve — merge as-is. No blocking issues; no changes applied.**

Clean, well-scoped, and it correctly leverages the #16 native-title plumbing as the carrier signal rather than adding new polling. Below is the verification trail and a few optional notes.

## The make-or-break check: does the marker actually match?

This feature silently never-badges if the shell's marker glyph differs from the WebUI's by even one codepoint, so I verified both byte-for-byte:
- **WebUI** (`hermes-webui` @ master, `static/ui.js:6279`): `document.title = '● ' + document.title` — the glyph is **U+25CF** (BLACK CIRCLE) + U+0020, prepended only when `activeSessionHasPendingPromptAttention()`. The base title is `<sessionTitle> — <name>` with an em-dash U+2014.
- **Shell** (`windows.rs` `ATTENTION_MARKER`): **U+25CF**. Exact match — not a lookalike (U+26AB/U+2022/emoji). The badge will light.

`split_attention("● Untitled — Hermes")` → `(true, "Untitled — Hermes")` → `clean_title` → `"Untitled"`. Traced and unit-tested.

## Trace findings (all correct)

- **Two-way tracking is sound because the marker toggle *is* a title change.** When a background popup appears, `syncTopbar` prepends `● ` → `document.title` changes → `on_document_title_changed` fires for that (background) webview → `set_tab_title(…, attention=true)`. On resolve, the marker is removed → another title change → `attention=false`. The native hook (from #15) firing for background webviews is exactly what makes this work with zero polling.
- **Per-tab isolation**: each tab webview has its own hook using its own `wv.label()`; `set_tab_title` updates only that tab's entry then `emit_tabs` re-renders the whole strip from the snapshot — so a background tab lights up while the active one is untouched. That's the issue's core ask.
- **Serialization**: `TabEntry` derives `serde::Serialize`, so the new `attention` bool flows through the existing `tabs-changed` payload to `shell.html` with no plumbing change. Confirmed the single `TabEntry` constructor (strip.rs:256) and the single `set_tab_title` caller (windows.rs:508) are both updated to the new field/signature.
- **macOS bonus**: the `else` branch now stores the marker-free `title`, so the stray `●` no longer leaks into the native macOS window title (it would have, since #15 sourced titles natively there too). Good catch.
- **Invariant #7 preserved**: no capability or IPC change — this rides the title hook, not EMIT.
- **shell.html**: dot is appended left of the title span, `aria-label="waiting for you"`, tooltip becomes `"… — waiting for you"`, `--hermes-fg` is a defined var, and `rgba()` (not `color-mix`) keeps it engine-version-agnostic.

## Verification

- `cargo fmt` / `clippy --all-targets -D warnings` / `cargo test` green (9 pass, +1 `splits_pending_attention_marker`); `cargo xwin check` (Windows target) clean.
- **Linux smoke on this branch** ([run 27516951363](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27516951363), green): the screenshot still renders the tab as "Smoke chat" (stripped) with no stray dot (fake server emits no marker → `attention=false`) — confirms the 4-arg `set_tab_title` + `split_attention` refactor didn't regress #16's title path on real WebKitGTK.
- **Rendered the strip markup** (exact shell.html CSS, dark-theme defaults) to visually confirm the one thing logic tests can't: the amber dot + 12%-opacity tint read clearly at a glance, sit cleanly left of the title, and don't clip the 38px strip. (Shared in the review thread on my side.)

## Optional notes (none blocking)

1. **Latent coupling: attention is gated on a non-empty clean title.** `apply_reported_title` runs `if clean_title(title).is_empty() { return; }` after `split_attention`, so a pending marker on a title that cleans to empty (e.g. `● — Hermes`) would drop the flag along with the empty title. This is effectively unreachable today — the WebUI always emits `<sessionTitle|"Untitled"> — <name>`, never an empty session title with a marker — so it's a note, not a bug. If you ever want to harden it, compute `attention` and apply it independent of the empty-title guard.
2. **Cosmetic squeeze**: the 7px dot + 6px gap eats ~13px of a min-width-72px tab; the title ellipsis absorbs it, but narrow attention tabs show less text. Acceptable.
3. **Active + attention**: an active tab keeps its control-bg (no amber tint) but still shows the dot — sensible (you're already looking at it). Intentional per the `:not(.active)` rule.

## Recommendation

Merge as-is. Correct, codepoint-verified against the live WebUI, regression-checked on real WebKitGTK, and the badge renders cleanly. The one remaining gate is the manual smoke the PR already calls out: trigger a tool approval in a *background* Win/Linux tab and confirm it lights amber and clears on answer — the automated smoke proves the path runs and the refactor is regression-free, but not a live pending-popup. Natural v0.3.6.

PR #20, opened 2026-06-16
Per-tab profile isolation: give each strip tab its own cookie jar (#3)

### LINKED ISSUE #3: Multi-tab: sidebar shows wrong profile after tab switch; profile switch intermittently drops that profile's default workspace

nesquena:
Full resolution of #3 — Windows/Linux multi-profile tab bleed (three independent reports: b3nw, Lemz, + b3nw's "not all active chats appear" follow-up).

## Root cause (confirmed, not a WebUI bug)

The WebUI selects the active profile **per-client via an HttpOnly `hermes_profile` cookie** (no query/header/path alternative — verified in the server). Separate browsers isolate correctly because they have separate cookie jars; the desktop strip's tabs **all shared one jar**, so switching profile in one tab flipped the cookie for every tab. That single root cause produces all three reported symptoms:
- sidebar/profile sticks to the last-loaded profile across tabs,
- other profiles' active chats are missing from the sidebar,
- the per-profile default workspace intermittently fails to apply.

## Fix (wrapper-side — the issue's proposed direction)

**Each tab webview now gets its own `data_directory`** (Windows/Linux) → its own cookie jar → profile selection is genuinely per-tab.

To avoid regressing the common single-profile / authenticated case (a fresh jar would mean "new tab = default profile + re-login"), **a new tab is seeded from the opener tab**:
1. Before creating the webview, read the opener's cookies for the target origin via the native cookie store (`cookies_for_url` — sees HttpOnly cookies), on the worker thread so Tauri's dispatcher marshals to the UI thread without the [documented main-thread `cookies()` deadlock](https://github.com/tauri-apps/wry/issues/583).
2. Create the tab at `about:blank` with its isolated `data_directory`.
3. `set_cookie` the seeded cookies, then `navigate()` to the target.

Seeding-before-navigation is **race-free**: WebKitGTK's `set_cookie` is synchronous (pumps until the add completes) and WebView2's `AddOrUpdateCookie` completes before `navigate`, so the first real request already carries the inherited `hermes_profile` + `hermes_session`. Net behavior: a new tab opens on your current profile and stays logged in, and only diverges when you switch *that tab's* profile — multi-profile tabs no longer bleed.

**Lifecycle:** partitions are session-scoped — removed on tab close (`close_tab`) and window close (`forget_window`), and wiped at startup (`clear_partitions` in `setup`, before any tab opens) so an orphan from a crash can't accumulate. (Per-close removal is best-effort since the OS may briefly hold the WebView2 folder; the startup wipe is the safety net.)

**Platform scope:** gated to `cfg!(not(target_os = "macos"))`. macOS uses native tabs (this module is Windows/Linux only) and WKWebView has no `data_directory`. macOS native-tab isolation would be a separate change via `data_store_identifier` (macOS 14+) and isn't part of issue #3.

## Known side effect (intentional)

`data_directory` isolates *all* per-origin storage, so **localStorage is now per-tab too** — e.g. the WebUI's theme/skin localStorage sync no longer crosses tabs. This is consistent with the "each tab is like a separate browser" model the reporters rely on; cookies (profile + login) are seeded, theme is not. Window chrome theme still seeds from the native theme cache.

## Verification

- `cargo fmt`, `cargo clippy --all-targets -- -D warnings`, `cargo test` (9) green.
- `cargo xwin check --target x86_64-pc-windows-msvc` clean — the isolation path is live on Windows and compiles cleanly there.
- All Tauri/wry APIs confirmed in the locked sources: `data_directory`, `cookies_for_url`/`set_cookie` (read+write, all three desktop engines), `navigate`, `PageLoadPayload::url()`.

**Verification gap — please review/smoke before release.** The isolation + seeding path only *executes* on Windows/Linux (`isolate` is `false` on macOS, so even `HERMES_FORCE_STRIP` won't exercise it on a Mac). The logic is reasoned and compiles clean on both targets, but the real behavior needs a live check on Windows/Linux:
1. Open two tabs, set them to different profiles → each sidebar shows its own profile; switching tabs doesn't cross-contaminate; both profiles' chats are visible in their own tab.
2. Single profile: open a new tab → it inherits the current profile and (if auth) stays logged in.
3. If auth is enabled, confirm the seed carries the session (no re-login on a new tab).

Suggested for the independent review: confirm the seed-before-navigate ordering holds on real WebView2 (the AddOrUpdateCookie→navigate sequencing), and that per-tab WebView2 user-data folders don't contend. The Linux smoke already exercises the isolated-tab create+load path.

Closes #3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## [v0.3.7]
+
+### Fixed
+
+- **Windows/Linux: multiple tabs on different profiles no longer bleed into
+  each other.** The WebUI scopes the active profile to a per-client HttpOnly
+  `hermes_profile` cookie, but every tab shared one cookie jar — so switching
+  profile in one tab flipped it for all of them: the sidebar stuck to the
+  last-loaded profile, other profiles' active chats went missing, and the
+  default workspace intermittently failed to apply. Each tab now gets its own
+  isolated data partition (its own cookie jar), so profile selection is
+  genuinely per-tab — the same isolation you get from separate browser windows.
+  A new tab is seeded with the opener tab's profile + login cookies, so it still
+  opens on your current profile (and stays logged in) and only diverges when you
+  switch it. Partitions are session-scoped and cleared on tab close and at
+  startup. (#3, reported by b3nw and Lemz.)
+
 ## [v0.3.6] — 2026-06-14
 
 ### Added
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -193,6 +193,9 @@ fn main() {
                 app.set_menu(m)?;
             }
             bridge::install(&handle);
+            // Wipe stale per-tab cookie partitions from a prior run before any
+            // tab opens (issue #3 — partitions are session-scoped).
+            strip::clear_partitions(&handle);
             {
                 use tauri_plugin_global_shortcut::GlobalShortcutExt;
                 // Conflict with the Swift app's identical default is expected
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -257,6 +257,15 @@ pub fn add_tab(app: &AppHandle, window_label: &str) {
         None
     };
     if let Some(ref dir) = partition {
+        // Guarantee a FRESH jar at the point of use. Tab labels are
+        // deterministic and recur every run (window_seq resets to 0 at
+        // startup, so the first tab is always `tab-1-1`), while both the
+        // per-close removal and the startup `clear_partitions` wipe are
+        // best-effort — the OS can hold the WebView2 folder open. If both ever
+        // fail, a same-named tab would otherwise inherit a prior session's
+        // cookies (stale profile/login) instead of opening on the default.
+        // Removing here makes session-scoping hold regardless of cleanup.
+        let _ = std::fs::remove_dir_all(dir);
         let _ = std::fs::create_dir_all(dir);
     }
     // Read the opener's cookies for the target origin BEFORE the new webview is
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -11,8 +11,9 @@
 use crate::state::{AppState, TabEntry, WindowTabs};
 use crate::{bridge, prefs, tunnel, windows};
 use serde_json::json;
+use std::path::PathBuf;
 use std::sync::atomic::Ordering;
-use tauri::webview::{Color, WebviewBuilder};
+use tauri::webview::{Color, Cookie, WebviewBuilder};
 use tauri::window::WindowBuilder;
 use tauri::{
     AppHandle, Emitter, LogicalPosition, LogicalSize, Manager, Webview, WebviewUrl, Window, Wry,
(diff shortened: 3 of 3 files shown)


nesquena:
# Deep review (end-to-end trace + source-level mechanism verification)

**Verdict: approve — merge after the one hardening commit I pushed to the branch (`088f8c9`).** Root cause is correct (confirmed not a WebUI bug), the fix is the right approach, and I verified the load-bearing mechanism at the source rather than taking the claims on faith. One real robustness gap — the kind the PR half-acknowledges — is now closed.

## Root cause & approach — confirmed

Verified against the issue's three reports and the WebUI: the active profile is a per-client HttpOnly `hermes_profile` cookie, and all strip tabs shared one jar. Per-tab `data_directory` (separate jar) is the correct wrapper-side fix, and it matches what separate browsers do (Lemz's "works across browsers, breaks in desktop" is the clean natural experiment).

## The mechanism, traced (not just the claims)

The risky parts are the async cookie seeding and the seed-before-navigate ordering, so I followed them through tauri/wry 2.11.2 / 0.55.1:

- **Threading / the wry#583 deadlock is genuinely avoided.** `cookies_for_url` dispatches `CookiesForUrl(url, tx)` and **blocks the caller on `rx.recv()`** — so the actual read (which pumps `gtk::main_iteration()` on WebKitGTK) runs on the event-loop thread while the worker waits. That's *why* `add_tab` must be off-main — and I confirmed **all four `add_tab` call sites are off-main** (the orchestrator, the `tab_new` command, the bridge `shortcut` handler, and `open_new_session`'s non-macOS `thread::spawn`). On macOS `open_new_session` runs inline on main, but `isolate` is a compile-time `cfg!(not(target_os="macos"))`, so the cookie code is fully inert there even under `HERMES_FORCE_STRIP`. No deadlock path exists.
- **Seed-before-navigate is race-free for a concrete reason.** `set_cookie` and `navigate` are both fire-and-forget `send_user_message` enqueues, so they're processed **FIFO on the main thread**: all `SetCookie` handlers run before the `Navigate` handler. And each commit is synchronous when its handler runs — WebView2 `AddOrUpdateCookie` is a synchronous COM call; WebKitGTK `set_cookie` pumps until `add_cookie` completes. So the cookies are in the jar before `navigate(target)` is even dispatched, and the first real request carries them. The PR's "synchronous" framing is right; the deeper guarantee is the ordered queue.
- **Cookie attributes survive the round-trip.** `cookie_into_win32` carries name/value/domain/path/expiry(max_age|expires)/HttpOnly/SameSite/Secure — so the seeded `hermes_profile` + `hermes_session` keep their scoping and are actually sent on the first request.
- **about:blank handoff is clean.** `navigation_allowed` lets `about:` through (engine-internal), the `on_page_load` `scheme == "about"` guard skips the pre-seed load so the anti-flash reveal only fires on the real target, and `clear_partitions` runs before `reconnect` opens any tab.

**Verification:** `fmt`/`clippy -D warnings`/`test` (9) green; `cargo xwin check` clean (the isolation path is live on Windows); **Linux smoke green** ([run 27642971372](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/27642971372)) with the screenshot confirming the isolated create → `about:blank` → `navigate(target)` path loads and renders on real WebKitGTK, title intact.

## Fix applied to the branch (`088f8c9`)

**Guarantee a fresh jar at creation, not just via best-effort cleanup.** Tab labels are deterministic and recur every run (`window_seq` resets to 0 at startup → the first tab is always `tab-1-1`), and the partition path is derived from the label. Both cleanup paths — per-close `remove_tab_partition` and the startup `clear_partitions` wipe — are explicitly best-effort because, as the PR notes, the OS can hold the WebView2 folder open. If *both* ever fail on a locked folder, the next same-named tab would reuse the stale partition and inherit a **prior session's profile/login** instead of opening on the default — silently defeating the session-scoped guarantee. The fix is a `remove_dir_all(dir)` immediately before `create_dir_all(dir)` in `add_tab`: the dir is never live for a brand-new unique label within a run, so removing it only ever clears a stale prior-run partition. This upgrades isolation from best-effort-twice to guaranteed at the point of use. (Re-ran all gates + smoke after — still green.)

## Notes (non-blocking)

1. **localStorage is now per-tab too** (PR-acknowledged, intentional) — the WebUI's cross-tab theme/skin sync stops crossing tabs; window chrome theme still seeds from the native cache. Consistent with the "separate browser per tab" model.
2. **`cookies_for_url` error → silent empty seed** (`.ok()…unwrap_or_default()`), which also fails open to the default profile + re-login. Fine as a fallback, but since the PR flags seeding as the thing to smoke-verify, a `log::debug!` on the `Err` branch would make "did the seed read succeed?" answerable from the log.
3. **A new *window* inherits the focused window's profile** via the `focused_active_webview` fallback (the first tab has no in-window opener). Consistent with the inherit model; worth a glance in the manual smoke to confirm it's the desired behavior for new windows.
4. **Disk**: each tab spawns a full webview data dir (cache/service worker/etc.); cleaned on close + startup, but heavy multi-tab sessions will grow `tab-partitions/` transiently.
5. macOS native-tab isolation (`data_store_identifier`, macOS 14+) is correctly out of scope for #3.

## Recommendation

Merge with `088f8c9`. The mechanism is verified end-to-end and the create+load path is green on real WebKitGTK. The remaining gate is the live multi-profile check the PR already calls out — CI can't stage an auth-enabled multi-profile server, so before publish: two tabs on two profiles (no sidebar bleed, each shows its own chats), a new tab inherits the current profile and stays logged in, and the auth session carries with no re-login. Natural v0.3.7.

PR #21, opened 2026-06-16
macOS per-tab profile isolation (#3 follow-up)

nesquena:
Follow-up to #3 / v0.3.7: fixes the profile-cookie bleed on **macOS**, which v0.3.7 didn't cover.

## Double-checking both platforms (as asked)

- **Windows/Linux (v0.3.7): believed fixed, runtime-unverified.** The strip gives each tab its own `data_directory` (separate cookie jar); the mechanism was traced at the source level in the #20 review and the Linux smoke confirmed the isolated tab creates/seeds/loads on real WebKitGTK. The one thing never run is a live *multi-profile* smoke (CI can't stage an auth+multi-profile server). I did **not** change the Win/Linux path here.
- **macOS: was NOT fixed — this PR fixes it.** v0.3.7's isolation was `cfg!(not(target_os = "macos"))`-gated, and macOS uses native window tabs (separate `WebviewWindow`s) that share one `WKWebsiteDataStore`. So switching profile in one tab switched the others — exactly the maintainer's report. Confirmed: `windows.rs` had zero isolation.

## The macOS fix

`WKWebView` has no `data_directory`, so isolation uses **`WebviewWindowBuilder::incognito`** (→ a fresh ephemeral `nonPersistentDataStore` per window, all macOS versions, auto-cleaned):

- **Each tab opened from an existing one** (`has opener`) → `incognito(true)` → its own ephemeral cookie jar. No shared store ⇒ no bleed.
- **Seeded from the opener**: read the opener's cookies for the target (native store API — sees HttpOnly), `set_cookie` them onto the new window, then `navigate(target)`. Loads `about:blank` first so the seed lands before the first real request (mirrors the strip's `add_tab`). So a new tab inherits the current profile + login and only diverges when switched.
- **The first window (no opener) keeps the default persistent store** — it's the only window ever using it, so nothing shares a jar, *and* login/profile still persist across restarts for the common single-window case (no re-login-every-launch regression).

### Threading (invariant #12)

The native cookie read/write pump the main run loop (`acceptInputForMode`). Doing that inside a tao menu/IPC callout — where macOS window creation runs inline — would re-enter `draw_rect` and self-deadlock (the v0.2–0.3.2 Cmd+T freeze class). So the seed is deferred to the **GCD main queue** (`macos::run_on_main_async`), which drains *between* callouts with the handler mutex free. Queued last, so it runs after the `addTabbedWindow` block. The `about:blank` pre-seed load is skipped in `on_page_load` so the anti-flash reveal only fires on the real target.

## Side effect (intentional, same as Win/Linux)

Isolated tabs use an ephemeral store, so their localStorage (theme/skin sync) is per-tab; window chrome theme still seeds from the native cache. The first window keeps full persistence.

## Verification

- `cargo fmt`, `clippy --all-targets -D warnings`, `cargo test` (9) green.
- `cargo xwin check` (Windows target) clean — the macOS-only seed/`seed_target` code is correctly `cfg`'d out.
- All Tauri/wry APIs confirmed in the locked sources: `incognito` → `nonPersistentDataStore` per webview (wkwebview/mod.rs), `cookies_for_url`/`set_cookie`/`navigate` on `WebviewWindow`, and `wait_for_blocking_operation` pumps the main runloop (so the GCD-deferred ordering is what keeps it safe).

**Verification gap — needs a macOS smoke (you can run this):**
1. Two tabs on two profiles → switching profile in one does **not** change the other; each sidebar shows its own profile + chats.
2. Open a new tab while in a profile → it inherits that profile and stays logged in.
3. No freeze on Cmd+T (the deferred-seed path) — `HERMES_TEST_TAB_AFTER=<secs>` drives the Cmd+T path and heartbeats the main thread if you want a scripted check.
4. The `open_browser: seeding … (N cookies)` debug log (Reveal Log File) confirms the seed ran.

Suggested for the independent review: confirm `incognito` yields a *distinct* `nonPersistentDataStore` per `WebviewWindow` (the load-bearing isolation assumption), and that the deferred seed can't race the first paint.

Refs #3

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,19 @@
 # Changelog
 
+## [v0.3.8]
+
+### Fixed
+
+- **macOS: multiple tabs on different profiles bled into each other** (the
+  v0.3.7 fix only covered the Windows/Linux tab strip). macOS uses native
+  window tabs, which shared one cookie store, so switching profile in one tab
+  switched the others too. Each tab opened from an existing one now gets its own
+  isolated (ephemeral) cookie store, seeded with the opener tab's profile +
+  login so it still opens on your current profile and only diverges when you
+  switch it — the same behavior the Windows/Linux fix gives. The first window
+  keeps the persistent store, so a single-window setup still remembers your
+  profile and login across restarts. (#3, reported by the maintainer on macOS.)
+
 ## [v0.3.7] — 2026-06-16
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@
   switch it — the same behavior the Windows/Linux fix gives. The first window
   keeps the persistent store, so a single-window setup still remembers your
   profile and login across restarts. (#3, reported by the maintainer on macOS.)
+- **A new tab now reliably inherits the current tab's profile.** The seed copied
+  the opener's cookies via a URL-filtered lookup, but that lookup drops
+  host-only cookies on macOS — and the WebUI sets the profile cookie host-only —
+  so a new tab fell back to the default profile instead of the one you were on.
+  Seeding now copies the opener's whole cookie store (a tab only ever loads one
+  origin), so the profile cookie transfers on every platform.
 
 ## [v0.3.7] — 2026-06-16
 
--- a/src-tauri/src/macos.rs
+++ b/src-tauri/src/macos.rs
@@ -59,6 +59,15 @@ fn dispatch_main_async(f: impl FnOnce() + Send + 'static) {
     };
 }
 
+/// Run a closure on the GCD main queue — drained by the runloop *between* tao
+/// callouts, so the handler mutex is free. Use this for main-thread work that
+/// pumps the run loop (e.g. the native cookie read/write, which spins
+/// `acceptInputForMode`): doing that inside a tao callout would re-enter
+/// `draw_rect` and self-deadlock the main thread (invariant #12).
+pub fn run_on_main_async(f: impl FnOnce() + Send + 'static) {
+    dispatch_main_async(f);
+}
+
 /// Attach `new_win` to `host`'s native tab group (Cmd+T behavior).
 pub fn add_tabbed_window(host: &WebviewWindow, new_win: &WebviewWindow) {
     let host2 = host.clone();
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -268,36 +268,45 @@ pub fn add_tab(app: &AppHandle, window_label: &str) {
         let _ = std::fs::remove_dir_all(dir);
         let _ = std::fs::create_dir_all(dir);
     }
-    // Read the opener's cookies for the target origin BEFORE the new webview is
-    // created. Safe here: add_tab runs on a worker thread, so Tauri's cookie
-    // dispatcher marshals to the UI thread without the main-thread deadlock the
-    // docs warn about. (`hermes_profile` is HttpOnly, but the native cookie
-    // store API sees it.)
+    // Read the opener's cookies BEFORE the new webview is created. Safe here:
+    // add_tab runs on a worker thread, so Tauri's cookie dispatcher marshals to
+    // the UI thread without the main-thread deadlock the docs warn about.
+    // (`hermes_profile` is HttpOnly, but the native cookie store API sees it.)
+    //
+    // Use cookies() (whole store), NOT cookies_for_url(): the latter filters by
+    // URL, and on WKWebView that filter drops host-only cookies entirely — the
+    // WebUI sets `hermes_profile` host-only (no Domain attribute), which is the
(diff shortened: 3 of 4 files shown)


nesquena:
# Deep review (source-level mechanism + live macOS verification)

**Verdict: approve — merge after a manual multi-profile smoke.** The core mechanism is correct and I verified it both at the source and live on real macOS, including the deadlock-sensitive path. I did **not** apply code changes: the committed code is sound, and the branch is under active edit (I can see a seed-logging improvement in flight), so I kept this review-only.

## Make-or-break: does `incognito` give a *distinct* store per window?

Yes — verified at the source, since the whole fix collapses if incognito windows share one store. `WebviewWindowBuilder::incognito(true)` → wry `wkwebview/mod.rs:235`: `WKWebsiteDataStore::nonPersistentDataStore(mtm)`. That's Apple's factory that returns a **new** ephemeral data store on every call (unlike the `defaultDataStore` singleton), invoked once per webview build. So each macOS tab-window gets a genuinely separate cookie jar ⇒ the HttpOnly `hermes_profile` can't bleed. Confirmed.

## Deadlock safety (invariant #12) — the riskiest part, verified at source *and* live

The native cookie read/write pump the main run loop: `cookies_for_url`/`set_cookie` → `wait_for_blocking_operation` → `NSRunLoop.mainRunLoop().acceptInputForMode_beforeDate(...)` in a loop (1s timeout). Pumping that **inside a tao callout** (where macOS window creation runs inline, holding tao's non-reentrant handler mutex) would re-enter `draw_rect` and self-deadlock — the v0.2–0.3.2 Cmd+T freeze class. The PR defers the seed to the GCD main queue (`run_on_main_async`), which drains *between* callouts with the mutex free, so the pump is safe. Queued after the `addTabbedWindow` block (GCD FIFO).

I exercised the exact deferred-seed Cmd+T path live on real macOS (`HERMES_TEST_TAB_AFTER`), which opens a seeded incognito tab and then heartbeats the main thread: **8/8 heartbeats fired on every run** (including the anomalous one below) — the main thread never froze. Invariant #12 holds.

## Seeding correctness

- **Race-free ordering**: within the deferred block, `cookies_for_url` → `set_cookie`(×N) → `navigate` run sequentially on the main thread, and each cookie write pumps until its completion handler fires, so all cookies are committed to the new ephemeral jar before the navigation request — a new tab's first request carries the inherited profile/login.
- **Attributes preserved**: `cookie_into_wkwebview` carries name/value/domain/path/max-age/expires/secure/HttpOnly, so the seeded `hermes_profile` + session keep their scoping ("stays logged in" is mechanically sound).
- **First-window-persistent invariant holds**: only the opener-less first window uses the persistent store, and macOS *hides* (never destroys) the last window, so no second window ever shares that jar while login/profile still persist across restarts for the single-window case.
- `about:blank` pre-seed load is correctly skipped in `on_page_load` (anti-flash reveal only on the real target), and `navigation_allowed` lets `about:` through.

## Live test results (real macOS, against a live server)

- **Deadlock-free**: every seeded-Cmd+T run kept the main thread responsive (8/8 heartbeats).
- **Correct steady state**: three clean runs each opened exactly one tab (`tabbed main-2 into main-1`), no anomalies.
- **One-off I could not reproduce** (reporting for transparency, not pinning on this PR): in a first run started *without* clearing prior app state (and where my SIGTERM didn't take, leaving a lingering process), extra tab-windows appeared over ~30s. It did **not** reproduce across three clean runs, and it never froze. I attribute it to leftover/zombie-process or window-restoration state from that dirty start rather than the seed code — but since I couldn't fully explain it, **watch for unexpected extra tabs during the manual smoke**; if it recurs on a clean launch, ping me and I'll dig back in.

**Other verification:** `cargo fmt`/`clippy -D warnings`/`cargo test` (9) green; `cargo xwin check` clean (the macOS-only `seed_target`/seed block correctly `cfg`s out). APIs confirmed in the locked sources (`incognito`, `cookies_for_url`/`set_cookie`/`navigate` on `WebviewWindow`, the runloop-pump).

## Notes (non-blocking)

1. **localStorage is per-tab on isolated macOS tabs** (intentional, same as Win/Linux) — the WebUI's cross-tab theme/skin sync won't cross native tabs; window chrome theme still seeds from the native cache.
2. **Seed reads have a 1s timeout** → on timeout the tab falls open to the default profile (re-login) rather than hanging — a good fail-open. The in-flight logging improvement on the branch (logging the seeded cookie names / the read-error branch) is exactly the observability that makes "did the seed carry?" answerable from the log — nice.
3. Ephemeral stores need no disk cleanup (WebKit auto-reclaims them), so there's no partition-lifecycle hole to defend here — cleaner than the Win/Linux disk-partition approach.

## Recommendation

Merge after the manual multi-profile smoke, which is the one thing neither CI nor a single-profile run can cover: two tabs on two profiles (switching one doesn't change the other; each sidebar shows its own profile + chats), a new tab inherits the current profile and stays logged in, and no freeze on Cmd+T (already confirmed deadlock-free here). Natural v0.3.8.

Issue #22, opened 2026-06-17
macOS: can't drag/move the window when only one tab is open


nesquena-hermes:
## Summary

> *"on macbook, I can't seem to drag/move the window around like I can with other apps, not sure if that is a program setting, kinda weird one."* (**b3nw**, macOS, v0.3.7)

On macOS, when a window has **only one tab**, you cannot drag it by the title bar area to move it. b3nw narrowed it down himself:

> *"when there is only 1 tab, you can't move it around"*
> *"if you have a tab [more than one], then the border shows and you can move it"*

So dragging works once the **native macOS tab bar** is visible (2+ tabs), but with a single tab there is no draggable region.

## Likely root cause

macOS windows are built with `TitleBarStyle::Overlay` + `hidden_title(true)` + `tabbing_identifier` (`src-tauri/src/windows.rs` ~L238-244). With the overlay titlebar, the draggable region is the **WebUI page's `.app-titlebar`**, made draggable by the bridge setting `data-tauri-drag-region` on it every 3s (`src-tauri/src/bridge.rs` `MACOS_TITLEBAR`, L72-82):

```js
var tb = document.querySelector('.app-titlebar');
if (tb && !tb.hasAttribute('data-tauri-drag-region')) tb.setAttribute('data-tauri-drag-region', '');
```

When 2+ tabs are open, macOS shows its **native NSWindow tab bar**, which is an AppKit-native draggable region independent of any web IPC — so dragging works. When there's a single tab, there's no native tab bar, and dragging falls back to the web `.app-titlebar` drag region. `data-tauri-drag-region` relies on the webview's JS→native drag IPC, which is unreliable in **remote-origin (External-URL) webviews** — the same webview-IPC limitation that broke the title `EMIT` bridge in #15 (closed). So the single-tab case has no working draggable surface.

The `.app-titlebar` element does exist in the current WebUI DOM (`static/index.html` L120), so the selector isn't the problem — it's the drag-region IPC not functioning on the remote webview.

## Files
- `src-tauri/src/windows.rs` (`TitleBarStyle::Overlay` window builder ~L238-244)
- `src-tauri/src/bridge.rs` (`MACOS_TITLEBAR` drag-region injection ~L60-83)
- relevant prior art: #15 (page→native bridge silently no-ops in remote webviews)

## Suggested direction
[...]

Issue #19, opened 2026-06-15
Drag-to-reorder tabs in the strip tab bar (Windows/Linux)


nesquena-hermes:
## Summary

> *"let me drag/drop to re-arrange them"* (**b3nw**)

Allow reordering the strip tabs by dragging them left/right, like a browser tab bar.

## Why it's not possible today

Tabs render in a fixed order from the backend snapshot. `renderTabs(snapshot)` in `src/shell.html` iterates `snapshot.tabs` in array order and appends a `.tab` div per entry; there is no drag handling anywhere (the only `data-tauri-drag-region` usages are for window dragging on the strip spacer/splash, not tab reordering). The order is `WindowTabs.tabs: Vec<TabEntry>` in `src-tauri/src/strip.rs`, and nothing mutates that order except push (add) and remove (close).

## Suggested shape

- **Frontend** (`src/shell.html`): make each `.tab` draggable (`draggable="true"` + `dragstart`/`dragover`/`drop`), or pointer-drag with a position calc; on drop, call a new `tab_reorder` command with the new index.
- **Backend** (`src-tauri/src/strip.rs`): add `reorder_tab(window_label, tab_label, new_index)` that moves the entry within `entry.tabs`, fixes up `entry.active` to follow the moved/displaced tab, then `emit_tabs` so all listeners re-render. Re-run `layout()` if needed (the active webview stays the same; only strip order changes).
- **Command** (`src-tauri/src/main.rs`): register the `tab_reorder` invoke handler.

## Notes

- The visible webview doesn't change on reorder — only the strip's display order and the stored `Vec` order. Keep the active tab pointing at the same `tab_label` after the move.
- Pairs naturally with tab persistence (companion issue): once order is user-controlled, the saved order is what should be restored.
- Separate from **#7** (rename tabs), which changes a tab's label, not its position.

nesquena-hermes:
Cross-platform confirmation from **b3nw** (Discord #desktop-apps, 2026-06-17, v0.3.7):

> *"interestingly, on the macbook I can re-arrange tabs, windows no such luck"*

[...]

Issue #18, opened 2026-06-15
Persist and restore open tabs across restart/upgrade (Windows/Linux strip mode)


nesquena-hermes:
## Summary

> *"small fr — can we not lose all our tabs when we upgrade? 😂"* (**b3nw**)

On Windows/Linux strip-tabs mode, restarting the app (including an upgrade) loses every open tab. The app reopens with a single fresh **"New Tab"** instead of the set the user had open.

## Why it happens

The strip tab registry is **runtime-only**. `AppState.strip` is a `Mutex<HashMap<String, WindowTabs>>` (`src-tauri/src/state.rs`), and `WindowTabs { tabs, active, tab_seq }` lives entirely in memory. `prefs.rs` persists connection settings, zoom, theme cache, and the first-window frame — but **never the tab list**. On window close, `strip::forget_window` drops the registry entry. On launch, `conn::reconnect` opens one window and seeds a single tab. Nothing ever reads tabs back.

So there is no code path that could restore tabs — they only exist for the lifetime of the process.

## What "restore" needs to capture

Each tab is a child webview pointing at `prefs.target_url` (`strip::add_tab` → `WebviewUrl::External(target)`); the *session* shown is the WebUI client's own state inside that webview. Tab labels (`tab-{winSeq}-{tabSeq}`) are regenerated every launch, so the WebUI's per-webview localStorage can't carry the selection across restart on its own.

A faithful restore should persist, per window:
- the **number of tabs** and their **order**
- each tab's **current URL** at save time (so a deep-linked WebUI session is reopened, not just the root)
- the **active** tab index

…then on boot recreate that many tabs navigating to the saved URLs instead of seeding one bare "New Tab".

## Files

- `src-tauri/src/state.rs` — `WindowTabs` / `TabEntry` (source of truth to serialize)
- `src-tauri/src/prefs.rs` — add a persisted `windowTabs`/`session` blob in `prefs.json` (tauri-plugin-store); save on tab add/close/select/navigate, debounced
- `src-tauri/src/strip.rs` — `open_browser_window` should restore saved tabs on first window open instead of seeding one; capture per-tab URL on navigation
- `src-tauri/src/conn.rs` — boot path that decides what to open

## Notes

- Interacts with **#9** (per-tab connection target) — once a tab can point at a chosen connection, that choice is part of what must persist.
[...]

Issue #9, opened 2026-06-12
Support multiple / per-tab server connections (e.g. live vs test instance)


nesquena-hermes:
## Summary
*"Being able to add different connections would be nice -- then have one tab for a live system and another connected to a test system."* (**Deor**; +1 from **b3nw**, 2 reactions)

Currently a single connection target is configured in prefs (`prefs.rs` `target_url`, `connection_mode`). Allow multiple saved connections and let a tab/window target a chosen one.

## Files
- `src-tauri/src/prefs.rs` (connection config -- currently single `target_url`)
- `src-tauri/src/conn.rs`, `src-tauri/src/strip.rs` (per-tab target)
- `src/prefs.html` (connection management UI)

nesquena-hermes:
Adding the requester's refined framing + code grounding.

**@b3nw** re-stated this concept in #desktop-apps (2026-06-17), proposing a specific UI shape:

> Can the desktop app enable more than 1 hermes-webui connection, and just have "groups" of tabs aligned to each server you're connected to.

So beyond "a tab can target a chosen connection," the ask is to **group the strip's tabs by their connection** — one visual cluster of tabs per connected server.

**What drove it (and an honest demand note).** b3nw wanted to run a Hermes agent on a MacBook for `computer_use` (its own hermes-webui) alongside his Docker gateway instance, all in one desktop window — connect one tab to the Docker box, another to the Mac. The original pain was that a remote gateway with an SSH execute setting messages fine, but *"the hermes-webui session still executes everything locally, not via the ssh setting."* He then found his case *"is actually working without a second hermes-webui"* and explicitly said *"I don't think I'm going to need it, but I think the concept is still valid."* → treat this as **concept-validated, nice-to-have**, not blocking.

**Grounding the gap (current master, v0.3.6):**
- `TabEntry` (`src-tauri/src/state.rs:17`) carries only `label` / `title` / `attention` — **no per-tab `url` or `connection` field**, so a tab can't address a different server today.
- `add_tab` (`src-tauri/src/strip.rs:167-180`) reads the **single global** `p.target_url` (+ `connection_mode`) from prefs for *every* new tab → all tabs in a window point at the same server by construction.
- `Prefs` (`src-tauri/src/prefs.rs:18-22`) stores one `target_url` + `connection_mode`, not a list of saved connections.

[...]

PR #24, opened 2026-06-17
Tabs & sessions (v0.4.0): restore windows/tabs, per-tab profile dot, drag-reorder, macOS single-tab drag fix

nesquena:
A meaty **tabs & sessions** release across macOS, Windows, and Linux. Closes the top of the roadmap's tab-management cluster: **#18, #8, #19, #22**.

### #18 — Restore windows & tabs across restart / in-app update  *(the headline)*
Quit (or auto-update + relaunch) and your workspace comes back: every window's tabs, their order, the active tab, each tab's URL, and each tab's **profile**.
- New `session.rs`: a small `Session { mode, target, windows[] }` blob in `prefs.json`. Capture is deduped (writes only on change), empty-guarded (never clobbers a real session with a transient zero-window moment), and reads on the main thread off a worker.
- **Win/Linux (strip):** `strip::session_windows` / `restore_window`; `add_tab` was split into `add_tab` + `add_tab_with` so restore reuses the exact seed→navigate path.
- **macOS (native tabs):** native `NSWindowTabGroup` is the source of truth (we get no KVO for drag-out/merge/reorder), queried via objc2 (`macos::session_windows`); `windows::restore_macos_window` rebuilds each group through the freeze-safe `add_tabbed_window` (invariant #12). Restored tabs are incognito + re-seeded with their saved profile selector.
- Restore runs **once per launch**, only when the saved session matches the current mode+target.
- **Caveats (documented):** only the non-sensitive `hermes_profile` selector is persisted — never login/auth cookies — so an authenticated server re-prompts after restart; a different server's session isn't restored.

### #8 — Per-tab profile dot (Win/Linux strip)
Each tab shows a color-keyed dot for its active profile (default profile = no dot); hover shows the profile name. Capitalizes on the v0.3.7/v0.3.8 per-tab isolation — `TabEntry` now carries the captured `hermes_profile`.

### #19 — Drag tabs to reorder (Win/Linux strip)
`strip::reorder_tab` + `tab_reorder` command + drag handlers in `shell.html`. The active tab follows its label across the move, and the new order is what gets saved/restored. (macOS reorders via native tabs already.)

### #22 — macOS: can't drag the window with only one tab
Switched window dragging to WebKit-native `-webkit-app-region: drag` on `.app-titlebar` (interactive children marked `no-drag`) instead of Tauri's JS drag region, which no-ops in remote-origin webviews — so it works regardless of tab count. Same remote-IPC family as the #15 title and #12 external-link fixes.

### #5 — deferred (noted in CHANGELOG "Known issues")
The icon's "white fringe" is a light border baked into the artwork; a clean fix needs the source logo + a proper macOS-template icon pass, not pixel surgery on the rendered PNGs.

## Safety / invariants
- **Invariant #12 (Cmd+T freeze):** restore reuses the GCD-deferred `add_tabbed_window`; cookie/URL reads stay outside tao callouts.
- **wry nil-URL panic:** `webview.url()` unwraps a nil `WKWebView.URL` on a not-yet-navigated webview, and capture can race one — guarded by `session::capture_url` (caught + skipped). Found and fixed during local restore testing.
- **Invariant #9:** all window/webview creation stays on worker threads (Win/Linux) / the existing macOS inline path.

## Testing (macOS, against a fake WebUI)
- Native-tab capture → saved `1 window / 2 tabs`; relaunch → `session: restoring 1 window(s), 2 tab(s)` → `tabbed main-2 into main-1`, **no panic**.
- Strip mode (`HERMES_FORCE_STRIP=1`) → restored 2 tabs, then `add_tab` → 3rd tab, **no freeze**.
- Cmd+T heartbeats **8/8**, `windows=2`, no freeze.
- `cargo test` (12, incl. new session serde/cookie tests) + `cargo clippy` clean + Windows `cargo xwin check` green.

Visual checks left for the maintainer on real hardware: the profile dots (#8), drag-to-reorder feel (#19), and single-tab window drag (#22).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,49 @@
 # Changelog
 
+## [v0.4.0] — 2026-06-17
+
+A tabs & sessions release across macOS, Windows, and Linux.
+
+### Added
+
+- **Your windows and tabs come back after you quit or update the app**
+  (issue #18). The shell now remembers each window's tabs — their order, which
+  one was active, each tab's URL, and each tab's profile — and restores them on
+  the next launch (including after an in-app update relaunches). Previously
+  every restart reopened a single empty tab and the rest were lost. Works on all
+  three platforms: macOS rebuilds the native window tab groups; Windows/Linux
+  rebuild the tab strip. Each restored tab reopens on the profile it was on (the
+  profile selector is re-seeded). Notes: a server that requires login will ask
+  you to sign in again after a restart (only the profile selector is persisted,
+  never login/auth cookies); the session is saved for the server you're
+  connected to and isn't restored if you switch to a different server.
+- **Per-tab profile dot** (issue #8, Windows/Linux tab strip). Each tab shows a
+  small colored dot keyed to its active profile, so several profiles open at
+  once are easy to tell apart at a glance; the default profile shows no dot.
+  Hovering a tab shows its profile name. Pairs with the per-tab profile
+  isolation added in v0.3.7/v0.3.8.
+- **Drag tabs left/right to reorder them** in the Windows/Linux tab strip
+  (issue #19). macOS already reorders via native window tabs; the custom strip
+  had no way to. The new order is what gets saved and restored.
+
+### Fixed
+
+- **macOS: you couldn't drag the window by the title bar when only one tab was
+  open** (issue #22). Window dragging relied on Tauri's JS-based drag region,
+  which doesn't work reliably in the remote-content webview — so with a single
+  tab (no native tab bar to grab) the window wouldn't move. Dragging now uses
+  WebKit's built-in native drag region (`-webkit-app-region`), which is
+  independent of that fragile IPC and works regardless of how many tabs are
+  open. (Same remote-webview limitation behind the earlier title (#15) and
+  external-link (#12) fixes.)
+
+### Known issues
+
+- The app icon still shows a faint light edge on its rounded corners (issue #5)
+  — the icon artwork has a light border baked in, so a clean fix needs the
+  source logo and a proper icon pass rather than editing the rendered images.
+  Tracked for a follow-up.
+
 ## [v0.3.8] — 2026-06-16
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.3.8",
+  "version": "0.4.0",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.3.8"
+version = "0.4.0"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.3.8"
+version = "0.4.0"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -59,13 +59,32 @@ const PASTE_SUPPRESS: &str = r##"
 
 /// S7/S8/S9 — macOS overlay-titlebar integration: traffic-light clearance,
 /// hide the page logo (collides with traffic lights), CSS rule for hiding the
-/// page titlebar when the native tab bar shows, drag region on .app-titlebar.
+/// page titlebar when the native tab bar shows, and a NATIVE drag region on
+/// `.app-titlebar`.
+///
+/// Window dragging uses WKWebView's built-in `-webkit-app-region: drag` (the
+/// same mechanism AppKit/Electron use for frameless drag) rather than Tauri's
+/// `data-tauri-drag-region`. The Tauri attribute drives dragging through a
+/// JS→native IPC that is unreliable in remote-origin webviews — the same
+/// remote-IPC fragility that broke titles (#15) and external links (#12) — so
+/// with a single tab (no native tab bar to drag) the window couldn't be moved
+/// at all (issue #22). `-webkit-app-region` is honored natively by WKWebView,
+/// independent of IPC/CSP/capability scope, and works regardless of tab count.
+/// Interactive children are marked `no-drag` so the titlebar's buttons/inputs
+/// still receive clicks. The `data-tauri-drag-region` attribute is kept as a
+/// belt-and-suspenders fallback (harmless when the CSS already handles it).
 const MACOS_TITLEBAR: &str = r##"
   try { document.documentElement.style.setProperty('--traffic-light-width', '80px'); } catch (e) {}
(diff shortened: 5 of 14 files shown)


nesquena:
# Deep review — ⚠️ request changes (reproducible macOS crash in session restore)

Big, well-structured PR and most of it is solid (details below). But I found a **reproducible crash on macOS** in the headline feature, so I'd hold merge until it's fixed. I caught it on the **first** live launch, not by reading — so it'll hit users.

## 🔴 Blocking: macOS SIGABRT when the autosave races a just-opened tab

**Repro:** built `c930496`, ran on macOS against a live server, opened a second tab. The app aborted. Log:

```
thread 'main' panicked at wry-0.55.1/src/wkwebview/mod.rs:1349: called `Option::unwrap()` on a `None` value      ← webview.url() on a nil WKWebView.URL
macos: tabbed main-2 into main-1
open_browser: seeding main-2 from main-1 — 0 cookie(s): []
thread 'main' panicked at tauri-runtime-wry-2.11.2/src/lib.rs:1678: called `Result::unwrap()` on an `Err` value: PoisonError { .. }
fatal runtime error: failed to initiate panic, error 3, aborting       ← SIGABRT (.ips written)
```

**Mechanism (traced to source):**
1. The new 4s session autosave → `capture` → `macos::session_windows` calls `member.url()` on the main thread for every content window.
2. A tab opened moments earlier is still on its pre-seed `about:blank` and its native `WKWebView.URL` is **nil**. wry's `url_from_webview` does `webview.URL().unwrap()` → **panic**.
3. `session::capture_url`'s `catch_unwind` *does* catch the unwind (the app continues — you can see it reach "tabbed"/"seeding")... **but the panic already poisoned a tauri-runtime-wry mutex** while it was held inside the message dispatch.
4. The deferred seed then calls `new_win.navigate()` → `lib.rs:1678` locks that poisoned mutex → `PoisonError` → unwrap → "failed to initiate panic… aborting" → **SIGABRT**.

So `capture_url` is guarding at the wrong layer: `catch_unwind` can't undo the in-wry lock poisoning. The crash races the ~4s tick against the gap between a tab's creation and its first navigation commit — small window, but it triggered on my first launch and will recur whenever you open tabs.

**Scope:** macOS-only. WebKitGTK's `url()` is `unwrap_or_default()` (safe) and WebView2's is `Result`-based; only wkwebview unwraps a nil `NSURL`. But macOS is your daily driver, and it's introduced by this PR (the autosave + `session_windows` + `capture_url` are all new here).

**Recommended fix:** never call `webview.url()` on a webview that hasn't committed a navigation. Track a per-tab/window "has loaded once" flag (set in `on_page_load` *after* the existing `about:` guard), and in both `session_windows` paths call `url()` only when loaded — otherwise fall back to the tab's saved/target URL. After the first commit the URL is non-nil **and** reflects `history.pushState` (the WebUI navigates sessions via pushState — `sessions.js:2774` — so the live `url()` is genuinely needed for fidelity; `payload.url()` from `on_page_load` would lose the session route). Keep `capture_url` as a backstop, but it can't be the primary guard. I held off applying this myself since it's a multi-file change touching files you're actively iterating on (`state.rs`/`strip.rs`/`windows.rs`) — happy to implement + re-run the repro to confirm if you'd like.

## Secondary

- **`maybe_restore` returns `true` without confirming a window was actually built.** `conn::run` then skips the fallback `open_browser`, so a total restore failure (every `restore_window`/`restore_macos_window` no-ops) leaves **zero** windows — and on Win/Linux a later Destroyed event could `maybe_quit_after_close`. Cheap guard: only return `true` if at least one window was created; else fall through to the bare-window open.
- **Recheck native-tab-group capture fidelity after the crash fix.** My one capture wrote a 2-tab group as **2 windows × 1 tab** — but that run crashed, and the 4s autosave can also capture in the gap before the GCD-deferred `addTabbedWindow` forms the group. Worth confirming a real 2-tab group round-trips as 1 window / 2 tabs (and that a capture racing the deferred attach doesn't persist a split layout).

## Verified clean (traced, not assumed)

- **Restore integration** (`else if !session::maybe_restore` — restore-or-open-one on first connect; first-connect gating makes later reconnects fall through to a normal open).
- **Invariant #12** preserved: both the profile-cookie capture (`capture_window_profile`) and the macOS restore (`add_tabbed_window` + seed/navigate) run via the GCD `run_on_main_async` deferral, outside tao callouts; the cookie reads pump the runloop only with the handler mutex free.
- **Cookie/profile round-trip**: `cookie_into_wkwebview` preserves domain/path/HttpOnly/expiry; `profile_cookie` reconstructs the host-only + HttpOnly + Path=/ selector (unit-tested). `cookies()` (not `cookies_for_url`) avoids the host-only-drop bug from v0.3.8.
- **Drag-reorder (#19)**: the JS `to -= 1` shift-compensation is consistent with the Rust `remove`-then-`insert` — I traced both directions and the index math lands correctly; clamp is panic-safe.
- **Profile dot (#8)**: stable hue hash, default profile = no dot, profile name in the tooltip; attention dot correctly takes precedence.
- **#22 drag**: `-webkit-app-region: drag` on `.app-titlebar` with interactive children `no-drag` is the right IPC-independent mechanism (same family as the #15/#12 fixes); confirm it actually moves the window on real WKWebView (your noted manual check).
- **Capture safety**: `capture` runs `capture_inner` on the main thread off a worker (blocks on a 2s timeout, never deadlocks); `persist` is `restoring`-guarded, empty-guarded, deduped, single-flight. `session.rs` serde round-trips + partial-blob defaults are tested.
- `cargo fmt` / `clippy --all-targets -D warnings` / `cargo test` (12) green; `cargo xwin check` (Windows) clean — the macOS-only restore/capture code `cfg`s out correctly.

## Recommendation

**Request changes** — fix the macOS `webview.url()`-on-nil crash before merge (it's the one thing standing between this and a strong v0.4.0), add the `maybe_restore` zero-window guard, and re-verify the native-tab-group capture. Everything else is in good shape. Ping me to land the URL-guard fix and re-run the crash repro.

nesquena:
Thanks — great catch on the SIGABRT; you traced it exactly right. Fixed in `fb110ae`.

## Blocking crash — fixed
You nailed the mechanism: `capture_url`'s `catch_unwind` caught the unwind but **couldn't undo the tauri-runtime-wry mutex poisoning** that happened while the panic propagated through the dispatch — so the deferred `navigate()` then aborted. Guarding at the wrong layer.

**Fix (as you recommended): never call `url()` on a webview that hasn't committed a navigation.** New `AppState.navigated` set, marked in each `on_page_load` *after* the existing `about:` guard, and gated in **both** `session_windows` paths — the live `url()` is read only for navigated webviews (so `history.pushState` session routes are still captured for loaded tabs), otherwise we fall back to the saved/target URL. `capture_url` stays as a backstop. Cleaned up on tab/window close (`forget_navigated`).

## Secondary — both addressed
- **Zero-window guard:** `maybe_restore` now returns `false` if restore built no windows, so `conn::run` falls through to a bare `open_browser` — a total restore failure can't leave the app window-less.
- **Native-tab-group fidelity:** the gate also fixes this — a window isn't captured until it's navigated, and the GCD `addTabbedWindow` runs (and the group forms) well before the network load commits, so capture always sees the formed group. Confirmed below.

## Re-verified live on macOS (slow fake server, 1.2s response, to widen the pre-navigation window)
- Cmd+T while the 4s autosave ticks repeatedly over the about:blank gap → **8/8 heartbeats, no panic, no SIGABRT, no new `.ips`**. (Same scenario that aborted on `c930496`.)
- Native 2-tab group captures **and** restores as **1 window / 2 tabs** (active preserved) — not a 2×1 split.
- `cargo fmt --check` (this was the red CI `test` check — fmt on the seed match arm), `clippy --all-targets -D warnings`, `cargo test` (12) all green.

Everything else from your "verified clean" list stands. Ready for re-review / merge once CI is green.

Issue #28, opened 2026-06-18
Restored tabs show 502 Bad Gateway on relaunch behind a proxy (readiness probe accepts 5xx; no per-tab retry)


nesquena-hermes:
## Summary
After the v0.4.0 session-restore feature (#18), closing and reopening the desktop app leaves the user with a broken workspace: tabs are reported as lost, **some restored tabs show `502 Bad Gateway`**, and the rest come back logged out.

Reported by **@b3nw** in #desktop-apps (2026-06-18), runs Hermes in Docker reached over Tailscale (i.e. a reverse proxy in front of the server):

> the app doesn't survive being closed/re-opened, tabs are lost, and some tabs return 502 bad gateways, the rest are in a logged out state

## Disposition of the three symptoms
1. **`502 Bad Gateway` on restored tabs** — real bug (root cause below).
2. **"the rest are in a logged out state"** — **by design**, not a bug. v0.4.0 deliberately persists only the non-sensitive `hermes_profile` *selector*, never auth/login cookies, so a server that requires login re-prompts after restart (CHANGELOG v0.4.0 says so explicitly). Calling it out so it isn't "fixed" by accident.
3. **"tabs are lost"** — most likely a consequence of #1: the tabs *are* rebuilt by restore but land on 502/error pages, so they look lost/broken rather than truly absent. (Could also be a restore-skip if the saved `target`/`mode` didn't match — secondary hypothesis, see Notes.)

## Root cause (current master, v0.4.1) — the 502
On launch, `conn::reconnect` gates the whole restore on a single reachability probe and then `maybe_restore` navigates every saved tab to its URL immediately:

- `health::http_reachable` **treats any HTTP response — including 4xx/5xx — as "reachable"; only a transport error fails** (`src-tauri/src/health.rs:3-14`). So when a reverse proxy (Tailscale/Cloudflare/nginx in front of a Docker upstream that is still booting) answers `502`, the probe returns `true`.
- With `ok == true`, `reconnect` proceeds to `session::maybe_restore` (`src-tauri/src/conn.rs:122-130, 146`), which rebuilds the windows and `strip::restore_window` re-navigates each tab to its saved URL (`src-tauri/src/strip.rs:218`, `session.rs:163-214`) **right away** — before the upstream behind the proxy is actually serving.
[...]

Issue #27, opened 2026-06-18
Can't drag files from the workspace tree to the composer (desktop only) — native drag-drop handler swallows in-page drag


nesquena-hermes:
## Summary
In the desktop app you can no longer drag an item from the **workspace file tree** onto the **chat composer** to reference it — the cursor shows the OS "no-drop" icon (red circle with a line through it) and the drop never registers. This works fine in a regular browser, so it's **desktop-only**.

Reported by **@b3nw** in #desktop-apps (2026-06-18):

> I used to be able to drag/drop from the workspace tree folders or items to reference them in the chat input area, now I'm getting an little red circle with a line through it … [confirmed] yea it's a desktop only thing

Nathan confirmed drag-and-drop still works on the latest WebUI build in a browser, which pins the problem to the wrapper.

## Root cause (current master, v0.4.1)
Tauri v2 / wry enables the **native OS file-drop handler by default**. When it's on, the native layer intercepts drag events over the webview before the page sees them, so the WebUI's own HTML5 `dragstart`/`dragover`/`drop` handlers never fire and the OS paints the "no-drop" cursor.

The WebUI fully implements internal tree→composer drag (verified on `nesquena/hermes-webui` master):
- File-tree rows are draggable and set the drag payload: `static/ui.js:13393-13396` — `el.setAttribute('draggable','true')` + `ondragstart` doing `dataTransfer.setData('application/ws-path', item.path)` / `'application/ws-type'`, `effectAllowed='copy'`.
- The composer is the drop target: `static/ui.js:13372` (`el.ondrop = async (e) => …`) with the `.composer-wrap.drag-over .drop-hint` affordance (`static/style.css:1835`).

None of the wrapper's webview builders disable the native handler:
- `WebviewBuilder::new(&tab_label, initial_url)` — `src-tauri/src/strip.rs:397` (strip tabs, Windows/Linux)
- `WebviewWindowBuilder::new(app, &label, WebviewUrl::External(initial_url))` — `src-tauri/src/windows.rs:210` (single-window / macOS)

`git log -S disable_drag_drop_handler --all` → **zero hits**: the handler has never been disabled, so in-page drag-drop has been suppressed on desktop since the wrapper shipped (b3nw's "used to be able to" was almost certainly in a browser).

## Suggested fix / fix shape
Disable the native drag-drop handler on the **content** webviews so the page's HTML5 drag-drop works:
[...]

Issue #26, opened 2026-06-17
Tab profile color dot is stale after switching profile inside an open tab (only updates on open/reload)


nesquena-hermes:
## Summary
The per-tab profile color dot (#8, shipped v0.4.0) only reflects the tab's profile **at the moment the tab is opened**. If you switch the profile *inside* an already-open tab, the dot keeps showing the old profile's color — so the only way to get the right color is to switch profile, open a **new** tab, then close the old one.

Reported by **@b3nw** in #desktop-apps (2026-06-17), Windows strip mode:

> so the tab icon colors is great, but they only reflect the profile when they're opened, so if you switch, you get the wrong color … which means you have to switch, then open a new one, then close the old one, to get a different color

(Screenshot: a "Macbook" tab with a red dot next to an "Ops" tab — the dot is stale relative to the in-tab profile.)

## Root cause (current master, v0.4.1)
`TabEntry.profile` (`src-tauri/src/state.rs:30`) is the value that drives the strip dot. It's only refreshed by `capture_tab_profile()` (`src-tauri/src/strip.rs:764`), which re-reads the tab's `hermes_profile` cookie from its isolated jar and repaints via `emit_tabs`.

`capture_tab_profile` is invoked from exactly **two** places:
1. The initial seed when the tab is created (`strip.rs:383`, the dot shown "before the first real load").
2. `on_page_load` with `PageLoadEvent::Finished` (`strip.rs:411-442`) — i.e. only on a **full page (re)load**.

Switching profile inside the WebUI updates the `hermes_profile` cookie and re-renders the SPA **without a full document load that the wry `on_page_load` hook fires for** — so `capture_tab_profile` never runs and `TabEntry.profile` stays stale. The dot is correct on open and on a hard reload, wrong after an in-tab switch. This matches b3nw's exact workaround (new tab picks up the fresh cookie at creation → correct color).

## Suggested fix / fix shape
Re-read the profile on the events an in-tab switch produces, not just full page loads. Options (cheapest first):

- **On tab activation:** call `capture_tab_profile` at the end of `select_tab` (`strip.rs:490-529`) — already repaints the strip, just add the cookie re-read on a worker thread. Catches the common "switch profile, click back to the tab" flow. Cheap, no WebUI change.
[...]

Issue #10, opened 2026-06-12
Reduce window chrome height: consolidate tab strip with the WebUI top bar, and/or make the tab bar toggle-able


nesquena-hermes:
## Summary
Related requests about vertical chrome space (Windows/Linux):
- *"The app tabs and the top bar in WebUI should be one line instead of two -- there's plenty of space to share."* (**Rod**) -- a screenshot showed the OS titlebar row + the tab-strip row stacked above the WebUI content.
- *"Tabs should be toggle-able (they throw off the balance of the design)."* (**Rod**)
- A Discord/explorer-style layout was suggested: window controls + menu inline with the tab row so the rows collapse toward one.

## Current state
On Windows/Linux the chrome is: OS titlebar + the 38px strip tab bar (`STRIP_HEIGHT = 38.0`, `strip.rs:21`) + the WebUI's own top bar. macOS already overlays the tab bar into the titlebar (`refresh_macos_chrome`).

## Proposed
- Option A: a setting to hide the strip tab bar when there's a single tab (or always), reclaiming 38px.
- Option B: integrate window controls / menu into the tab row and reduce stacking.

## Files
- `src-tauri/src/strip.rs` (`STRIP_HEIGHT`, layout)
- `src/shell.html` / `src/shared/shell.css`

nesquena-hermes:
**Follow-up from v0.6.1 testing (@b3nw, #webui, 2026-06-19).**

The hide-tab-bar toggle works:

> #10 yes [it works]

…but there's a discoverability trap — once the bar is hidden, the **only** way to bring it back is the `Ctrl+Shift+B` hotkey:

> but it also means if you don't know the control+shift+b hotkey, you're lost forever

So a user who hides the tab bar (or hits the hotkey by accident) and doesn't know the shortcut has no visible affordance to restore it.

**Suggested residual fix (within this issue's scope):** give the un-hide a discoverable path —
- keep a "Show tab bar" item available in the app menu / ⋯ menu even while the bar is hidden, and/or
- show a one-time hint/toast on hide ("Tab bar hidden — press Ctrl+Shift+B to show it"),
- and surface the `Ctrl+Shift+B` shortcut in the menu item label.

Leaving open to track the discoverability piece.

PR #29, opened 2026-06-18
Bug-squash (v0.5.0): file drag #27, restored-tab 502 + login persistence #28, profile dot #26, + rename #7 / hide-bar #10 / what's-new #6

nesquena:
A bug-squash release targeting the things that work **worse than the web**, plus tab/session breakage and a few tab conveniences. Picks up the cluster of issues filed since v0.4.x.

### #27 — Drag files from the workspace tree → composer (the headline)
Desktop left wry's **native OS drag-drop handler** enabled, which intercepts drag events over the page so the WebUI's own HTML5 drag-drop never fires (you got the "no-drop" cursor). Disabled it on the content webviews (`strip.rs`, `windows.rs` ×2). Bonus: also restores **Finder/Explorer file-drop-to-upload** (drops now reach the WebUI). The shell intercepts no native drops, so nothing is lost.

### #28 — Restored tabs come back as `502 Bad Gateway` behind a proxy
On launch the app treated *any* HTTP response — including a proxy's `502/503/504` while its upstream booted — as "ready," so restore reopened tabs onto gateway-error pages that never recovered. New `health::http_ready` rejects gateway 5xx (a real `401` login / `404` / `405` still counts as up — refines invariant #2 for *readiness* only); restore waits for the server to actually serve; the health loop treats 5xx as down and reloads content on recovery. Also a lenient `target` match (trailing slash) so a Tailscale URL diff doesn't silently skip restore.

### #28 (login persistence, Windows/Linux)
The "logged out after restart" half of #28. Each tab's on-disk cookie jar is now **reused** when its tab is restored instead of wiped — so login + cookies survive quit/reopen and updates, like a browser. Only the jar is reused; **nothing sensitive is written to prefs**. Partitions are removed on explicit tab close; startup keeps session-referenced jars and sweeps orphans. macOS tabs use incognito (can't persist) and still re-prompt. (`TabEntry.partition` / `SessionTab.partition`; `add_tab` refactored to a `TabSpec`.)

### #26 — Profile dot stale after an in-tab profile switch
It only refreshed on open / full reload, so switching profile *inside* a tab left the old color (and the first tabs after re-login had no dot). Now re-reads on tab activation + a light 4s sweep.

## Enhancements

- **#7 Rename tabs** — double-click the title (Win/Linux strip). Sticks regardless of page title, restored across restart, clears back to the page title.
- **#10 Hide the tab bar** (Windows) — ⋯ menu or **Ctrl+Shift+B** (also brings it back); content fills the window. macOS = native tabs; Linux gated off pending the GTK child-webview geometry fix (constraint #1).
- **#6 In-app version + What's New** — bundled changelog section in a small window + menu items (version was already in the ⋯ menu).

## Already shipped (v0.4.0/v0.4.1), closing alongside
#8 (profile dot), #18 (restore tabs), #19 (drag-reorder), #22 (macOS window drag).

## Verified
- macOS: builds, `cargo fmt`, `clippy --all-targets -D warnings`, `cargo test` (13, incl. new session/changelog tests). Windows `cargo xwin check` clean.
- Strip boots + restores through the refactored `TabSpec`/partition path with **no panic**; readiness **rejects a permanent 502** (error window) and a real server connects; session persists the new `partition`/`custom_title` fields.
- Left for real-hardware/CI confirmation: the actual file-drag on the live WebUI (#27), Win/Linux cookie-jar persistence (#28b is a `data_directory` runtime behavior), and the rename/toggle UX clicks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,47 @@
 # Changelog
 
+## [v0.5.0] — 2026-06-17
+
+A bug-squash release focused on things that worked worse than the web, plus tab
+and session fixes and a few tab conveniences.
+
+### Fixed
+
+- **You can drag files from the workspace tree onto the chat composer again**
+  (issue #27, macOS/Windows/Linux). The desktop wrapper left wry's native OS
+  drag-drop handler enabled, which intercepted drag events over the page so the
+  WebUI's own drag-drop never fired — you got the "no-drop" cursor. The handler
+  is now disabled on the content webviews, so in-page drag works **and**
+  dragging a file in from Finder/Explorer drops onto the composer to upload.
+- **Restored tabs no longer come back as `502 Bad Gateway` behind a reverse
+  proxy** (issue #28). On launch the app treated *any* HTTP response — including
+  a proxy's `502`/`503`/`504` while its upstream was still booting — as "ready,"
+  so it reopened your tabs onto gateway-error pages that never recovered.
+  Readiness now rejects those gateway statuses (a real login page / `404` / `405`
+  still counts as up), waits for the server to actually serve, and reloads any
+  tab if the server recovers while the app is open.
+- **Login now survives a restart on Windows/Linux** (issue #28). Each tab's
+  cookie jar is kept on disk and reused when its tab is restored, so you stay
+  logged in across quit/reopen and updates — matching a browser. (Only the jar
+  is reused; nothing sensitive is written to the prefs file. macOS tabs use an
+  isolation model that can't persist, so they still re-prompt for login.)
+- **The per-tab profile dot updates when you switch profile inside a tab**
+  (issue #26, Windows/Linux). It previously only refreshed when a tab was opened
+  or fully reloaded, so an in-tab profile switch left the dot showing the old
+  color; it now re-reads on tab activation and a light periodic sweep, and the
+  first tabs after a re-login get their dots too.
+
+### Added
+
+- **Rename a tab** by double-clicking its title (issue #7, Windows/Linux strip).
+  The name sticks regardless of the page title and is restored across restart;
+  clear it to fall back to the page title.
+- **Hide the tab bar** to reclaim its space (issue #10, Windows) — from the ⋯
+  menu or **Ctrl+Shift+B** (which also brings it back). macOS uses native tabs;
+  Linux is pending an upstream webview-geometry fix.
+- **What's New** in the app menu / ⋯ menu (issue #6) shows the current version
+  and this release's changelog in-app, with a link to the full changelog.
+
 ## [v0.4.1] — 2026-06-17
 
 ### Fixed
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.4.1",
+  "version": "0.5.0",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.4.1"
+version = "0.5.0"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.4.1"
+version = "0.5.0"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
--- a/src-tauri/capabilities/shell.json
+++ b/src-tauri/capabilities/shell.json
@@ -2,8 +2,8 @@
   "$schema": "../gen/schemas/desktop-schema.json",
   "identifier": "shell-windows",
   "description": "Full IPC for bundled shell pages (splash, error, preferences, tab strip)",
-  "windows": ["splash", "error", "prefs", "main-*"],
-  "webviews": ["splash", "error", "prefs", "strip-*"],
+  "windows": ["splash", "error", "prefs", "whatsnew", "main-*"],
+  "webviews": ["splash", "error", "prefs", "whatsnew", "strip-*"],
   "permissions": [
     "core:default",
     "store:default",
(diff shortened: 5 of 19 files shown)


nesquena:
# Deep review — ⚠️ request changes (one data-loss / profile-bleed bug; rest is solid)

Strong bug-squash PR and I verified the headline mechanics hold (details at the end). I found one real correctness bug — code-confirmed, not a crash — in the new login-persistence path that can both **wipe a tab's login** and **reintroduce the #3 profile bleed**. Worth fixing before merge since it undercuts the two guarantees this PR is strengthening (#28 persistence + #3 isolation).

## 🔴 Partition-name collision can wipe a restored tab's jar and make two tabs share one

**The mechanism (traced through `strip.rs`):**
- New tabs key their partition dir on the tab label: `tab-{window_seq}-{tab_seq}`, and `add_tab_with` does `remove_dir_all(dir)` then `create_dir_all(dir)` for a fresh jar (correct — my #20 fix, preserved).
- Restored tabs **reuse the saved partition name** (`partition_override`, e.g. `tab-1-3`) and skip the removal (correct — keeps the jar so login persists).
- But `tab_seq` **restarts at 0** on each launch, and `restore_window` doesn't advance it past the reused suffixes. So a restored tab can be *labeled* `tab-1-2` while *reusing* partition `tab-1-3`.

**Repro (Win/Linux):**
1. Open 3 tabs (partitions `tab-1-1/2/3`), **close the first** (or reorder) → window now has 2 tabs reusing `tab-1-2` and `tab-1-3`. Quit.
2. Relaunch → restore builds 2 tabs labeled `tab-1-1`, `tab-1-2`, reusing `tab-1-2`, `tab-1-3`. `tab_seq` is now **2**.
3. Open one new tab → `tab_seq` → **3** → label/partition `tab-1-3` → `remove_dir_all(tab-partitions/tab-1-3)` — which is the **live jar of the restored tab labeled `tab-1-2`**.

Result: that tab's login/cookies are wiped mid-session, and now two tabs point at `tab-1-3` → shared cookie jar → the exact profile bleed #3/#28 fixed. The pristine "never closed/reordered" case is safe (labels match reused names), but closing a non-last tab — then restart + open a tab — is a normal sequence. macOS is unaffected (incognito, no on-disk partitions); I couldn't live-repro because the partition path is `cfg`'d off on macOS, so this is a code-trace — please confirm on Win/Linux.

**Fix:** decouple new-tab partition ids from the reused namespace. Simplest: in `restore_window`, after adding the tabs, set
```rust
entry.tab_seq = entry.tab_seq.max(highest numeric suffix among that window's reused partition ids);
```
so new-tab labels never land on a live reused jar. (Or uniquify new-tab partition ids — a persisted counter / uuid — which would also let you retire the #20 remove-before-create entirely.) Happy to implement + have you verify on Win/Linux — I held off since it's a design choice on a branch you're actively iterating and I can't exercise the partition path on macOS.

## Verified solid

- **`http_ready` (#28)** is a sound, narrow refinement of invariant #2: only `502/503/504` are rejected; `401`/`403`/`404`/`405`/`500`/`501` still count as ready (a login page / present-but-erroring server is "up"), transport errors are down. Wired into direct connect + the health loop (5xx→down, reload-on-recovery) + the recovery loop. The lenient trailing-slash target match is a nice touch.
- **The #24 `url()`-on-nil crash fix landed** exactly as recommended: `session::has_navigated()` now gates every `webview.url()` read, and `mark_navigated` is set in `on_page_load` after the `about:` guard (with `forget_navigated` on close). That was the prerequisite for this PR's restore to be safe, so good to confirm it's solid.
- **Partition reuse-vs-fresh** logic is otherwise correct: reuse → `create_dir_all` only (keep jar); new → `remove_dir_all`+`create` (fresh). `TabEntry.partition` stores the stable id so it round-trips across multiple restarts. The startup `clear_partitions` keep-set (session-referenced names survive, orphans swept) is the right approach.
- **#27 drag-drop** — `disable_drag_drop_handler()` is on the 3 content webviews only (`strip.rs`, `windows.rs` ×2), not the shell/splash/prefs, so the page's HTML5 dnd fires and nothing native is intercepted. Correct fix.
- `cargo fmt` / `clippy --all-targets -D warnings` / `cargo test` (13) green; profile recapture sweep is Win/Linux-only with marshaled cookie reads.

## Secondary (non-blocking)

- **SSH-tunnel 502 gap:** the readiness fix is direct-mode only — the tunnel readiness probe (`tunnel.rs:115`) still uses `http_reachable`, so an SSH-tunneled proxy returning 502 while it boots can still pass tunnel-up and restore onto error pages. Consider `http_ready` there too (or document the limit).
- **`maybe_restore` still returns `true` without confirming a window was built** (carry-over from #24) — a total restore failure leaves zero windows with no fallback `open_browser`.
- **Health recovery reloads *all* content** (`eval_all_content("location.reload()")`) on any down→up — a transient 5xx blip would interrupt in-flight streams in tabs that weren't stuck. Consider reloading only tabs actually showing a gateway error, or gating on a sustained-down threshold.

## Recommendation

Request changes — fix the partition-name collision before merge (it silently breaks the login persistence and profile isolation this PR is meant to deliver), and ideally close the SSH 502 gap. Everything else is in good shape and the crash-fix carry-over from #24 is confirmed solid. Ping me to land the `tab_seq` fix and verify on Win/Linux.

nesquena:
Thanks — the partition-name collision is a real one, fixed in `1925af8`.

## Blocking: partition-name collision — fixed
You traced it exactly. `tab_seq` restarts at 0 each launch while restored tabs reuse higher partition suffixes (with gaps from closed/reordered tabs), so a later new tab could be labeled onto — and `remove_dir_all` — a restored tab's live jar.

Fix: `restore_window` now advances the window's `tab_seq` past the **highest reused partition suffix** before any new tab can be added, so new-tab names always clear the reused range. Extracted `partition_suffix()` + unit test.

**Verified on macOS forced-strip** (the `tab_seq`/label logic runs cross-platform even though the partition dir is `cfg`'d off): seeded a session whose one tab reuses partition `tab-1-5`, restored it, then opened a new tab → it came up as **`tab-1-6`**, not a colliding `tab-1-5`/lower. (Win/Linux is where the dir actually exists; the naming logic is identical, so the collision can no longer occur there.)

## Secondary — SSH 502 gap: fixed
`tunnel.rs` readiness now uses `http_ready` too, so a reverse proxy on the remote answering 502/503/504 while it boots no longer passes tunnel-up. Same gate as direct mode now.

## Already handled / accepted
- **`maybe_restore` zero-window guard** — it's already there (`session.rs`): if restore builds no windows it returns `false` and the caller opens a bare window. (Landed with the #24 fixes; flagging since the review listed it as outstanding.)
- **Health reload-on-recovery reloading all content** — kept as a deliberate trade-off: wry exposes no per-tab HTTP status, so we can't target only the gateway-errored tabs, and the 30s health cadence already debounces a brief blip (the server must be down across a poll boundary to trigger). The reported bug (launch-restore onto 502) is fixed by the readiness gate, not the reload. Open to tightening later if it proves noisy in the field.

`fmt` / `clippy --all-targets -D warnings` / `cargo test` (14) green. Ready for re-review / merge.

Issue #33, opened 2026-06-18
Windows: clicking the "⋯" overflow menu hangs the app (AppHangB1) — stuck always-on-top, Preferences/Quit dead, must kill from Task Manager


nesquena-hermes:
## Summary
On Windows v0.5.0, clicking the **"⋯" overflow menu** (top-right of the tab strip) hangs the app. The menu freezes and becomes non-interactable, the window gets stuck in an **always-on-top** state that blocks other apps from coming to the front, and **Preferences/Settings and Quit/Exit stop working** — the app has to be killed from Task Manager. **Reproducible.**

Reported by **@b3nw** in #desktop-apps (2026-06-18), Windows, after upgrading to v0.5.0:

> the 3 dots froze, and now they won't re-open
> I hit the 3 dots top right, it froze became non-interactable, and then was gone, after which the 3 dots was unresponsive. The app still worked, but settings and exit didn't work. Had to kill from task manager to get it back up.
> and it reproduces, after clicking on it, it seems to get stuck in some type of always ontop mode, which then stops other apps from being brought up above it? weird windows things I've never seen before.

(Screenshot shows the expanded overflow menu: New Tab, New Window, Hide Tab Bar, Reload, Find in Page, Zoom, Preferences, Open in Browser, v0.5.0, What's New, Check for Updates, Reveal Log File, Quit.)

Windows recorded it as an application hang:

```
The program hermes-webui-desktop.exe version 0.5.0.0 stopped interacting with Windows and was closed.

Fault bucket, type 0
Event Name: AppHangB1
P1: hermes-webui-desktop.exe
P2: 0.5.0.0
P3: 6a3363ad
P4: 77ec
P5: 67246080
```

`AppHangB1` = the main UI thread stopped pumping its message loop (a classic Windows main-thread block), which lines up exactly with the observed symptoms (frozen modal-looking menu, window stuck topmost/input-capturing, native menu actions like Preferences/Quit no longer firing).

## Root cause (strong hypothesis — current master, v0.5.0)
The overflow menu is popped up **synchronously on the main thread** from an IPC command:

- `src/shell.html:395` — the `#menubtn` click does `hermes.invoke("strip_menu", { window: WIN })`.
[...]

Issue #32, opened 2026-06-18
Bridge browser notifications to native OS notifications in the desktop app (Web Notifications API absent in webview)


nesquena-hermes:
## Summary
Browser notifications don't work in the desktop app. The WebUI front-end already implements native browser notifications (response-complete, approval-required, clarify-needed), but inside the desktop wrapper's embedded webview the Web Notifications API isn't available, so every notification silently no-ops.

Reported by **@b3nw** in #desktop-apps (2026-06-18); confirmed not-yet-working and flagged as next:

> **b3nw:** should browser "notifications" work in the desktop app?
> **Nathan:** Not yet but it's a good reminder I should work on that next!

## What the WebUI already does (so the desktop side is the only gap)
The web app drives the standard Web Notifications API from `static/messages.js`:
- `requestNotificationPermission()` → `Notification.requestPermission()` (~5303)
- `sendBrowserNotification()` / `_showPwaNotification()` → `new Notification(...)` and, when a service worker is active, `reg.showNotification(...)` (~5282, ~5325)
- Fired on: **response complete** (`'Response complete'`, ~3512), **approval required** (~3160), **clarification needed** (~3168), plus a "Send test" path.

So a normal browser tab gets desktop notifications today. The functionality is all there — it just needs the webview to expose (or the shell to bridge) the Notifications API.

## Why it fails in the wrapper
Tauri's webviews don't implement the W3C Web Notifications API the way a full browser does:
- **macOS (WKWebView):** no Web Notifications API — `'Notification' in window` is false, so `requestNotificationPermission()` hits the `notifications_unsupported` branch (~5304) and `sendBrowserNotification()` early-returns at `if(!('Notification' in window)) return;` (~5328).
- **Windows/Linux (WebView2 / WebKitGTK):** Web Notifications are commonly unavailable or unsurfaced in an embedded webview, so permission requests and `new Notification(...)` either throw or no-op.

Net effect: the user enables notifications in WebUI settings (the toggle appears to work), but nothing ever fires in the desktop app.

## Suggested fix / fix shape
Bridge the page's notification intent to native OS notifications via **`tauri-plugin-notification`** (already the canonical Tauri path; handles macOS Notification Center / Windows toasts / Linux libnotify, and the OS permission prompt).

Two routes (pick per reliability — note the remote-webview `EMIT` caveat from #15):
[...]

Issue #31, opened 2026-06-18
Profile color dot: missing on tab creation until another tab opens, and adopts wrong (first tab's) color after a session switch


nesquena-hermes:
## Summary
The per-tab profile color dot (#8, v0.4.0) has failure modes **beyond** the in-tab-switch staleness already tracked in #26. From @b3nw's testing on the current build (~v0.5.0, #desktop-apps 2026-06-18), the dot is also **missing at creation** and **wrong (adopts another tab's color) after a session switch**:

**Mode A — dot doesn't paint on a freshly created tab until you open *another* tab**
> new tab + change profile, no dot change … the first newly created tab doesn't update until you create a new tab
> new tab + change profile + new tab = both tabs update to their correct profile dots

So creating a tab and switching its profile leaves it dot-less; only the act of creating a *subsequent* tab triggers a repaint that finally colors the first one. (Related: earlier b3nw also saw "the first two tabs created after relogin also don't get colored dots" — same no-dot-on-create class, on the relogin path.)

**Mode B — dot adopts the *first* tab's color after a session switch**
> new tab + change profile + change session = wrong profile color dot (it adopts the first tab) instead of aligning to the new profile

Here the new tab doesn't just stay stale — it paints the **wrong** profile's color (the first tab's), so it's a wrong-*value* bug, not just a missing/stale one.

b3nw also notes the behavior is **timing-flaky** ("it appears sometimes there is a delayed update", "I've also seen that use the wrong colored dot"), consistent with a repaint that only fires on certain events.

## Relationship to #26
#26 is scoped narrowly to "dot is **stale** after switching profile **inside an already-open tab**; only updates on open/reload," with the root cause being `capture_tab_profile` firing only on tab-create seed + full page load. The modes here are adjacent but distinct and may not be fixed by the #26 in-tab-switch fix alone:
- **Mode A** = the *create-time seed* itself isn't painting/emitting until a later `emit_tabs` (triggered by opening another tab) flushes it. Initial seed/emit ordering bug.
- **Mode B** = wrong attribution — the new tab's dot resolves to the first tab's profile after a session switch. Possibly a stale index into the tabs vector during `emit_tabs`, or the seed reading the wrong cookie jar.

[...]

Issue #30, opened 2026-06-18
Tab restore is shallow: reopened tabs lose their session, title, and profile dot (only count survives)


nesquena-hermes:
## Summary
After the tab-persistence work (#18, shipped v0.4.0) and the 502 fix (#28), closing and reopening the desktop app now **preserves the tab count**, but each restored tab comes back **blank/fresh** — the session it was on, the tab title, and the profile color dot are all gone. So restore is *shallow*: you get N empty "New Tab" webviews instead of the workspace you had.

Reported by **@b3nw** in #desktop-apps (2026-06-18), testing the current build (~v0.5.0):

> X and re-open preserves the tabs, but wipes out the dots, titles, and sessions they were on
> … I still have 3 tabs, just they're basically reset to fresh — so kinda net-net there

So the count survives, the *content* of each tab does not. Net-neutral for the user: the tabs reappear but you still have to manually re-navigate each one back to its session.

## Why it happens (hypothesis — current master)
This is the exact gap #18 called out as required for a faithful restore but that the implementation doesn't yet cover. #18's own spec:

> A faithful restore should persist, per window: the number of tabs and their order, **each tab's current URL at save time (so a deep-linked WebUI session is reopened, not just the root)**, and the active tab index.

The current restore appears to persist **tab count (+ order)** but **not each tab's live URL** — so on boot every tab is recreated navigating to `prefs.target_url` (the root), not to the deep-linked WebUI session URL the tab was showing at save time. Landing on the root means:
- **session lost** — the WebUI opens its default/empty session, not the one the tab was on;
- **title lost** — the tab title is derived from the loaded session, so a root load shows the generic label;
- **profile dot lost/stale** — the dot is seeded from the tab's cookie/profile at creation; a bare root recreate doesn't carry the prior per-tab profile (see also the dot-specific issues #26 and the create/session-switch dot bug).

`TabEntry` (`src-tauri/src/state.rs`) is the serialize source; the per-tab `target`/current-URL is the field that needs to be captured at save and replayed at restore instead of seeding `target_url`.

## What a complete restore needs
- Capture each tab's **current URL** (the deep-linked WebUI session), not just the count — replay that URL per tab on boot.
- Restore the **active tab index** so focus lands where the user left it.
- Ensure the per-tab **profile** is restored so the color dot repaints correctly (ties into #26 and the dot create/session-switch bug).
- Title then follows naturally once the tab loads its real session.

## Scope
[...]

PR #34, opened 2026-06-18
Bug-squash (v0.6.0): Windows ⋯-menu hang #33, deep-session restore #30, profile dot #31, native notifications #32

nesquena:
A bug-squash release grabbing the cluster filed since v0.5.0 — including a Windows hang blocker.

## #33 — Windows: clicking the "⋯" menu hangs the app *(blocker)*
The strip menu was popped **synchronously inside the IPC command**, so on Windows the modal `TrackPopupMenu` loop wedged the WebView2 message pump (`AppHangB1`): the menu stuck, the window got stuck always-on-top, and Preferences/Quit stopped firing — only a Task Manager kill recovered it. `strip_menu` now marshals the popup onto the **main event loop** (`run_on_main_thread`) so its modal loop pumps normally and the command returns immediately. (This was the one remaining main-thread-blocking IPC command — the off-thread invariant #9 now covers it.)

## #30 — Restore was shallow (tabs came back blank)
Reopened tabs preserved only the **count** — each came back fresh (session, title, profile dot gone). Root cause: `webview.url()` doesn't reliably reflect the WebUI's in-app (SPA `pushState`) routing — notably on WebView2 — so restore replayed the **root** URL. Fix: a route reporter has the page `EMIT` its live `location.href`; `AppState.tab_urls` stores it; session capture prefers it over `url()`.

**Verified end-to-end on macOS:** capture saved `…/c/sess-abc?notif=true` (the deep route, not root), and on relaunch the server received `GET /c/sess-abc?notif=true` — the tab reopened on its exact session.

## #31 — Profile dot missing on create / wrong color after a session switch
Follow-on to #26. The dot is now re-read the moment the page navigates (a profile/session switch reports a route), via `recapture_tab_profile` — so it paints on a freshly created tab without needing to open another, and stops adopting a sibling tab's color after a session switch. Combines with the existing `select_tab` + 4s sweep capture paths.

## #32 — Browser notifications now fire natively
The embedded webview has no Web Notifications API, so the WebUI's notifications (response complete / approval required / clarification needed) silently no-op'd. Replaced the old permission-`denied` stub with a `window.Notification` **shim** that routes through the bridge to `tauri-plugin-notification` — so they appear in Notification Center / Windows action center / Linux tray. Honors the app's notifications pref. (Also removed the old crude DOM-mutation "response is ready" guess — the shim delivers the WebUI's real, specific notifications.)

**Verified:** `notif=true` round-tripped through the shimmed page (`'Notification' in window && permission === 'granted'`), and `new Notification(...)` reached the native bridge.

## Verification
- macOS: builds, `cargo fmt`, `clippy --all-targets -D warnings`, `cargo test` (14). Windows `cargo xwin check` clean.
- #30 + #32 verified live on macOS (forced-strip). #31's recapture fires on the proven route signal. #33 is a clean Windows-specific fix — **please field-verify on Windows** that the ⋯ menu opens/closes and the window stays interactable (Preferences/Quit fire) afterward.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,40 @@
 # Changelog
 
+## [v0.6.0] — 2026-06-18
+
+A bug-squash release: a Windows hang blocker, plus restore/notification/profile-dot fixes.
+
+### Fixed
+
+- **Windows: clicking the "⋯" overflow menu no longer hangs the app** (issue #33).
+  The menu was popped synchronously inside the IPC command; on Windows that
+  modal `TrackPopupMenu` loop wedged the WebView2 message pump (AppHangB1) — the
+  menu stuck, the window got stuck always-on-top, and Preferences/Quit stopped
+  working, forcing a kill from Task Manager. The popup now runs on the main
+  event loop so its modal loop pumps normally and the command returns at once.
+- **Reopened tabs come back on the session they were on, not a blank page**
+  (issue #30). Restore replayed the root URL, so each tab returned fresh
+  (session, title, and profile dot all gone). The page now reports its live URL
+  — including in-app (SPA) navigation that `WebView2`/`WKWebView` don't expose
+  through the wrapper — so restore reopens the exact deep-linked session.
+- **The per-tab profile dot updates right after you switch profile** (issue #31,
+  follow-on to #26). The dot is now re-read the moment the page navigates (a
+  profile/session switch), so it paints correctly on a freshly created tab
+  without needing to open another tab, and no longer adopts a sibling tab's
+  color after a session switch.
+
+### Changed
+
+- **Browser notifications now fire as native OS notifications** (issue #32).
+  The embedded webview doesn't implement the Web Notifications API, so the
+  WebUI's notifications (response complete, approval required, clarification
+  needed) silently did nothing. The app now bridges `window.Notification` to
+  native notifications, so they appear in Notification Center / the Windows
+  action center / the Linux notification tray. Honors the app's notifications
+  preference. (Replaces the old DOM-heuristic "response is ready" guess with the
+  WebUI's real, specific notifications.)
+
+
 ## [v0.5.0] — 2026-06-17
 
 A bug-squash release focused on things that worked worse than the web, plus tab
--- a/README.md
+++ b/README.md
@@ -1,18 +1,13 @@
 # Hermes WebUI Desktop
 
-The cross-platform desktop app for [hermes-webui](https://github.com/nesquena/hermes-webui)
-— run your Hermes agent's web UI as a real desktop app on **Windows, Linux, and macOS**,
-with one-time SSH tunnel setup for remote servers, native tabs, theme-matched window
-chrome, clipboard image paste, and background-response notifications.
+The cross-platform desktop app for [hermes-webui](https://github.com/nesquena/hermes-webui) — run your Hermes agent's web UI as a real desktop app on **Windows, Linux, and macOS**, with one-time SSH tunnel setup for remote servers, native tabs, theme-matched window chrome, clipboard image paste, and background-response notifications.
 
 <img width="1466" height="920" alt="image" src="https://github.com/user-attachments/assets/a7b81abc-f053-4e11-b2a4-ca3d4d8d72ea" />
 
-Built with [Tauri 2](https://tauri.app) (system webview + Rust — ~11 MB, no bundled
-browser). It is the cross-platform sibling of
-[hermes-swift-mac](https://github.com/hermes-webui/hermes-swift-mac), rebuilt 1:1 from
-its architecture (see [issue #555](https://github.com/nesquena/hermes-webui/issues/555)).
-On macOS, hermes-swift-mac remains the flagship app; this build exists for Windows and
-Linux users and for anyone who wants identical behavior across all three platforms.
+Built with [Tauri 2](https://tauri.app) (system webview + Rust — ~11 MB, no bundled browser). It is the cross-platform sibling of
+[hermes-swift-mac](https://github.com/hermes-webui/hermes-swift-mac), rebuilt 1:1 from its architecture (see [issue #555](https://github.com/nesquena/hermes-webui/issues/555)).
+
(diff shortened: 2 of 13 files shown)


nesquena:
# Deep review — approve, with three field-verification items

Solid bug-squash. I traced all four changes and found **no blocking bug** — the code is correct as written, and each change is a net improvement (no regressions). The open risks are all remote-webview behaviors I can't exercise from macOS, so they're field-checks rather than code defects. Two nice confirmations up front: my **#29 partition-collision fix landed** (`tab_seq = tab_seq.max(max_suffix)` in `restore_window`), and the **#24 `url()`-on-nil crash guard is intact** — capture now prefers the reported URL and only falls back to `url()` when `has_navigated`, so no crash regression.

## Per change

**#33 — Windows ⋯-menu hang.** Correct fix: deferring the modal `Menu::popup` to `run_on_main_thread` lets the IPC command return immediately (unblocking the WebView2 pump) and runs `TrackPopupMenu`'s modal loop on the main event loop where it pumps normally. The closure re-fetches the window/rebuilds the menu, so a closed window is handled.
- ⚠️ *Caveat (not blocking):* `run_on_main_thread` is the pattern invariant #12 warns against on macOS — a modal loop inside a tao callout can re-enter `draw_rect` and self-deadlock. It's safe here only because `strip_menu` is Windows/Linux-only (macOS uses native tabs); it's reachable on macOS solely via `HERMES_FORCE_STRIP` (dev). If the strip ever ships on macOS, this popup should go through the GCD `dispatch_main_async` path like `add_tabbed_window`.

**#30 — deep session restore.** Clean: the route reporter wraps `pushState`/`replaceState` + `popstate`/`hashchange` (with a 2s backstop) and `EMIT('route', location.href)`; `report_url` stores it in `tab_urls` (ignoring non-http so a pre-seed blank can't clobber a real route); capture prefers it over `url()`; `forget_url` is wired on close. The `url()` fallback stays `has_navigated`-gated, so the #24 crash guard holds.

**#31 — profile dot.** Re-reading the profile on each route report (`recapture_tab_profile`, spawned off the bridge callout) is the right trigger and composes with the `select_tab` + 4s sweep paths. Fine.

**#32 — native notifications.** The `window.Notification` shim → `EMIT('notify')` → `tauri-plugin-notification` is well-formed (permission `granted`, `requestPermission` resolves granted, inert instance methods), gated on the notifications pref, and the WebUI self-gates on `document.hidden` so it won't fire while a tab is focused. Strictly better than the old `denied` stub.

## Field-verification items (can't test from macOS — please confirm before relying on these)

1. **#30 and #32 both ride `EMIT` on remote content — the exact channel #15 found unreliable on WebView2.** The #15 capability fix (`http://*`) should make `EMIT` work for the configured origin, and you verified #30 live on *macOS*, but the WebView2 path is unconfirmed. Worth a Windows check that route reports arrive (deep restore reopens the exact session, not root) and that notifications fire. If `EMIT` is dropped on WebView2, #30 silently degrades to root-only restore (no worse than before, just not the fix).

2. **#32's real path may bypass the shim.** The WebUI's `_showPwaNotification` (messages.js) prefers `ServiceWorkerRegistration.showNotification` when an active SW exists and only falls back to `new Notification` (your shim) if that **rejects**. If the embedded webview has an active service worker whose `showNotification` resolves but displays nothing, real notifications get swallowed and never reach the shim. The PR verified the shim works when `new Notification` is called *directly* — not the WebUI's SW-first path. Please confirm a **real** WebUI notification (response complete / approval) actually surfaces natively; if it doesn't, also shim `ServiceWorkerRegistration.prototype.showNotification` (or have it route through the bridge). Related: removing `NOTIFY_WATCHER` means that on any platform where the SW swallows notifications, the new code notifies *less* than before (the old crude heuristic at least fired "response is ready") — so this is the one place with a small regression risk if the shim path isn't reached.

3. **#33** — the stated manual check: on Windows, the ⋯ menu opens/closes and the window stays interactable (Preferences/Quit fire) afterward.

## Recommendation

Approve. No blocking issues; the prior-review fixes (#24 crash guard, #29 collision) are confirmed in place. The three items above are field-verifications of remote-webview behavior, not code changes — #1/#2 in particular determine whether #30/#32 fully deliver on Windows, so worth a quick real-device pass before the release notes promise them. Happy to help wire a `registration.showNotification` shim if #32's SW path turns out to swallow notifications.

nesquena:
Thanks — approve received, and I took the one actionable item.

## Acted on: #32 SW-first path (`e183c3f`)
You're right that `_showPwaNotification` prefers `ServiceWorkerRegistration.showNotification` and only falls back to `new Notification` if the SW path *rejects* — so an embedded SW that resolves-but-shows-nothing would swallow notifications before the shim sees them. I now shim **`ServiceWorkerRegistration.prototype.showNotification`** too, routing it through the same `EMIT('notify')` → `tauri-plugin-notification` path (and resolving). So both the direct-`Notification` and SW-first paths fire natively — which also closes the small `NOTIFY_WATCHER`-removal regression you noted (no platform ends up notifying *less* than before).

## Noted / accepted
- **#33 `run_on_main_thread` vs invariant #12** — added a comment explaining it's safe here because the strip (and this command) is Windows/Linux-only; macOS reaches it only via dev `HERMES_FORCE_STRIP`, and a context-menu popup doesn't force the window redraw that triggers the `draw_rect` re-entrancy. If the strip ever ships on macOS, this goes through `dispatch_main_async`.
- **#1 EMIT-on-WebView2** — same channel the strip's shortcut/theme bridge already uses successfully on Windows post-#15 (the `http://*` capability fix), so I expect route reports + notifications to arrive there; flagged for the Windows pass below. If `EMIT` were dropped, #30 degrades to root-only restore (no worse than before).

## Field-verification (Windows, before relying on these)
1. ⋯ menu opens/closes; window stays interactable, Preferences/Quit fire (#33).
2. Quit/reopen reopens each tab on its **exact session**, not root (#30 — route EMIT arriving on WebView2).
3. A **real** WebUI notification (response complete / approval) surfaces natively (#32 — now covers the SW path too).

Proceeding toward cutting v0.6.0.

PR #36, opened 2026-06-19
fix: fully resolve ⋯-menu hang #33 (guard cookies() at chokepoint) + restored-tab 404 #37 — v0.6.1

### LINKED ISSUE #33: Windows: clicking the "⋯" overflow menu hangs the app (AppHangB1) — stuck always-on-top, Preferences/Quit dead, must kill from Task Manager

nesquena-hermes:
Fixes #33 (the part v0.6.0 / #34 didn't resolve).

## Status: the hang is NOT fixed on v0.6.0
@b3nw tested v0.6.0 after updating and the freeze persists:
> freeze is not fixed … it seems to trigger as you go over the version ID … **no, it's a time thing, 2-3 seconds after the 3 dots is open** … doesn't matter what you hover/select

This is exactly the failure mode predicted on #33: #34 marshaled the popup onto the event loop, but that only changed *where the popup is invoked from* — the native modal loop still owns the main thread while the menu is up, so the real trigger was never addressed.

## Root cause (the other half of the deadlock)
1. `Menu::popup` runs a nested Windows **`TrackPopupMenu` modal message loop** that owns the **main thread** until the menu is dismissed.
2. A periodic **4-second timer** (`main.rs`) calls:
   - `session::persist` → `capture` → reads each tab's **URL**, and
   - `strip::recapture_profiles` → `capture_tab_profile` → reads each tab's **cookie** (`wv.cookies()`).
   
   Both **marshal back onto the main thread** via the runtime dispatcher (the code already documents this: *"cookie reads marshal … must not be called from inside a wry callout"*).
3. When that marshal lands while the popup's modal loop already holds the main thread, it **re-enters WebView2 from inside the modal loop** — which WebView2 forbids — and the UI thread deadlocks (`AppHangB1`). ~2-3s after opening = the next 4s tick landing mid-menu, independent of hover/select. That's why "it's a time thing."

## The fix
A `menu_open` flag on `AppState`, set for the duration of the popup. The 4s timer **skips its webview reads while a menu is open** and catches up on the next tick once it closes. `recapture_profiles` also self-guards at the function level. Both paths are idempotent and only act on a real change, so a skipped tick is harmless — the dot/session state is captured on the very next tick.

Three layers:
- `strip_menu` sets/clears `menu_open` around `menu.popup` (`main.rs`, inside #34's event-loop closure).
- The 4s timer early-`continue`s while `menu_open` is set (`main.rs`).
- `recapture_profiles` early-returns while `menu_open` is set (`strip.rs`).

This is additive on top of #34 — keeps the popup on the event loop *and* closes the timer-marshal race.

## Scope
Desktop-wrapper only. Windows-specific symptom; guard is safe cross-platform (on macOS/Linux `popup` returns immediately so the flag is naturally short-lived). WebUI side needs nothing.

## Testing
- `cargo fmt --check` passes.
- `AppState` is only built via `AppState::new()` (updated); no struct-literal construction elsewhere; existing unit tests don't touch it.
- Full `cargo clippy -D warnings` / `cargo test` / platform builds run in CI (macos-14) — local `cargo check` can't run in this sandbox (no GTK system libs), so CI is the authoritative compile gate.
- **Windows verification wanted:** open the "⋯" menu, leave it open >5s, dismiss — confirm the window stays interactable, is **not** stuck topmost, and Preferences/Quit still fire; then confirm the profile dot still refreshes within ~4s after an in-tab profile switch. (@b3nw offered a Windows VM to remote into — that would let us confirm directly before release.)

Bumps to **v0.6.1** + CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Changelog
 
+## [v0.6.1] — 2026-06-19
+
+### Fixed
+
+- **The "⋯" overflow menu no longer hangs the app on Windows** (issue #33,
+  follow-up to v0.6.0). v0.6.0 moved the popup onto the event loop, but the hang
+  persisted: testers still saw the app freeze ~2-3 seconds after opening the
+  menu — the window stuck on top of everything, Preferences/Quit dead, only a
+  Task Manager kill recovered it. The popup runs a native modal loop that owns
+  the main thread until dismissed; meanwhile a periodic 4-second timer (session
+  autosave + the per-tab profile-dot refresh) reads each tab's cookie/URL, and
+  those reads marshal back onto the main thread. With the menu's modal loop
+  already holding the main thread, that re-entry deadlocked the UI — independent
+  of which item you hovered or selected, exactly matching the "it's a time thing"
+  report. The timer now skips its webview reads while a menu is open and catches
+  up on the next tick once it closes, so the menu stays responsive.
+
 ## [v0.6.0] — 2026-06-18
 
 A bug-squash release: a Windows hang blocker, plus restore/notification/profile-dot fixes.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [v0.6.1] — 2026-06-19
 
+A crash-and-bug-fix release.
+
 ### Fixed
 
 - **The "⋯" overflow menu no longer hangs the app on Windows** (issue #33,
@@ -9,13 +11,22 @@
   persisted: testers still saw the app freeze ~2-3 seconds after opening the
   menu — the window stuck on top of everything, Preferences/Quit dead, only a
   Task Manager kill recovered it. The popup runs a native modal loop that owns
-  the main thread until dismissed; meanwhile a periodic 4-second timer (session
-  autosave + the per-tab profile-dot refresh) reads each tab's cookie/URL, and
-  those reads marshal back onto the main thread. With the menu's modal loop
-  already holding the main thread, that re-entry deadlocked the UI — independent
-  of which item you hovered or selected, exactly matching the "it's a time thing"
-  report. The timer now skips its webview reads while a menu is open and catches
-  up on the next tick once it closes, so the menu stays responsive.
+  the main thread until dismissed; meanwhile background work that reads a tab's
+  cookie or URL marshals back onto the main thread, and that re-entry while the
+  modal loop is up deadlocked the UI — independent of which item you hovered or
+  selected, matching the "it's a time thing" report. Every such read — the
+  periodic 4-second autosave + profile-dot refresh, **and** the
+  navigation-driven profile re-read and page-load capture — now pauses while a
+  menu is open and resumes the moment it closes, so the menu stays responsive.
+- **A restored tab no longer shows "Session not available in web UI." on
+  launch** (issue #37, new in v0.6.0's deep-session restore). Behind a reverse
+  proxy the restored session's first load could 404 before the server was ready
+  to resolve it, leaving the tab on the empty-state until you switched away and
+  back. A restored tab that gets bounced to the home screen now retries its
+  saved session once, a couple seconds later — mirroring that switch-away-and-back
+  recovery — so the session you left off in comes back on its own. (If a tab
+  still lands on the empty-state behind a slow proxy, a manual reload or session
+  switch recovers it.)
 
 ## [v0.6.0] — 2026-06-18
 
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.6.0"
+version = "0.6.1"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -173,8 +173,17 @@ fn strip_menu(app: tauri::AppHandle, window: String) {
         if let Ok(menu) = menu::build_strip_menu(&app) {
             let state = app.state::<AppState>();
             state.menu_open.store(true, Ordering::SeqCst);
+            // Reset on scope exit, even if `popup` were to panic — a stuck
+            // `menu_open` would silently freeze session autosave + the profile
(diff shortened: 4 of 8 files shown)


nesquena:
## Deep review + hardening (applied directly to this branch, `eca6887`)

Reviewed against the goal of *fully* resolving #33, traced the Windows deadlock to its source, found a remaining gap, and batched in #37 for the v0.6.1 crash/bug-fix release.

### #33 — the timer guard alone wasn't complete; hardened at the chokepoint
The deadlock is **any synchronous `cookies()` read landing on the main thread while the popup's `TrackPopupMenu` modal loop owns it** — not just the 4s timer. Confirmed in wry 0.55:

```
cookies() → cookies_inner → webview2_com::wait_with_pump(rx)
```

`wait_with_pump` runs a **nested message pump** waiting for the async `GetCookies` result; inside `TrackPopupMenu`'s modal loop that hangs the UI thread (AppHangB1, window stuck topmost). (`url()` is a synchronous `Source` read — no pump — so it's safe; I guard it anyway as belt-and-suspenders.)

**The gap:** the PR guarded the 4s timer + `recapture_profiles`, but the **`route` bridge handler** also reaches `cookies()` via `recapture_tab_profile → capture_tab_profile`. A background (SSE-driven) navigation while the menu is open would still hang — almost certainly the "still crashing after the fix" path. `on_page_load → capture_tab_profile` was likewise unguarded.

**Fix:** guard at the **chokepoints** — `capture_tab_profile` (cookies) and `capture` (url) check `menu_open` — so *every* caller is covered (route handler, page-load, tab activation, the sweep). Audited all `cookies()` sites: the only other one (`add_tab`'s opener seed) can't run mid-menu — selecting "New Tab" dismisses the menu first. Also wrapped `menu_open` in a `Drop` guard so a panic in `popup` can't leave it stuck (which would silently freeze autosave + the dot).

### #37 — restored tab "Session not available" on launch (batched in)
A restored deep-link tab can 404 on its first `GET /api/session` behind a proxy before the server resolves it; the WebUI bounces to root and shows the empty-state ("switch away and back" recovers it). `TabSpec` gains `restore_retry` (the saved deep URL, set only for non-root restores); ~2.5s after create, if the tab has bounced to root, it re-navigates once to the deep URL — mirroring that recovery, with no penalty when the tab loaded fine. **Verified on macOS forced-strip:** bounce detected (`reported_url` = root) → deep re-navigation issued.

### Windows-correctness (can't run Windows; verified statically)
- Traced the actual wry WebView2 code: `cookies()` is the **only** `wait_with_pump` op, and it's now guarded on every reachable path → the nested-pump-inside-modal-loop deadlock can't occur.
- `cargo xwin check` (x86_64-pc-windows-msvc) is clean; local build + `clippy --all-targets -D warnings` + `cargo test` (14) green.
- The route `EMIT` that #37's retry relies on already works on WebView2 (post-#15 capability fix; same channel the strip's shortcut/theme bridge uses on Windows).

### Field-verification (Windows) before release
1. Open the ⋯ menu, leave it open >5s, hover/select around, dismiss → window stays interactable, not stuck topmost, Preferences/Quit fire (#33). The SSE-route case is now covered too.
2. Quit/reopen → the restored tab lands on its session (not "Session not available") (#37).

Issue #38, opened 2026-06-19
Windows: rename-tab dialog (#7) appears then vanishes instantly on double-click (v0.6.1)


nesquena-hermes:
## Summary

The rename-tab feature (#7, shipped v0.5.0) is broken on **Windows v0.6.1**: double-clicking a tab title briefly opens the rename window/dialog, then it **vanishes instantly** before you can type.

## Reporter

@b3nw, #webui, 2026-06-19, fresh **Windows v0.6.1**:

> #7 no, I can see it trying to bring up a rename window but it vanishes instantly, will see if I can get a video

## Notes

- The base feature itself landed in v0.5.0 (PR #29) — this is a Windows-specific failure of that implementation, so filing as a focused bug rather than reopening the #7 enhancement.
- Signature (popup appears then immediately dismisses, regardless of interaction) points at a **focus/lifecycle race** on the native rename input window: the dialog/overlay is created and then loses focus or gets dismissed on the same tick (e.g. a click-outside / blur handler firing from the same double-click event, or the modal being created off the main thread and torn down). Compare the ⋯-menu hang family (#33) for the Windows webview/main-thread interaction patterns.

## Suggested next step

Repro on Windows, instrument the rename-window create→focus→dismiss sequence; the fix is most likely to defer/guard the dismiss handler so the originating double-click doesn't immediately close the freshly-opened input.

Cross-ref: #7 (feature).

nesquena:
Fixed in **v0.6.2** (shipped in v0.6.3). Root cause: the active tab's `document.title` churns constantly while a chat streams, and each change rebuilt the entire strip — destroying the freshly-opened rename `<input>`. Two-part fix: `set_tab_title` now emits `tabs-changed` only when the rendered title/attention actually changes (`strip.rs`), and `renderTabs` never rebuilds while a rename input is open, releasing the rename only if the tab is closed mid-edit (`shell.html`). The rebuild-destroys-input mechanism is platform-independent, so the fix applies on Windows too.

Issue #37, opened 2026-06-19
v0.6.0: restored tab shows "Session not available in web UI." on launch (session exists — switch away/back loads it)


nesquena-hermes:
## Summary
New on **v0.6.0** (after the deep-session-restore work, #30/#34): the session a tab restores into on launch shows **"Session not available in web UI."** — the WebUI empty-state ("What can I help with?") with that message at the bottom. The session is **not actually gone**: switching to another session and back makes it load fine.

Reported by **@b3nw** in #desktop-apps (2026-06-19), Windows, v0.6.0:

> also, this is new — "Session not available in web UI."
> yea whatever session you land in when the app opens basically
> *(switch and switch back it's fine?)* yep

So it's specifically the **boot-time restore load** that fails; the same session loads correctly once you navigate away and back within the same app run.

## Disposition
The session exists server-side and is reachable a moment later (the switch-away-and-back recovery proves it), so this is **not** data loss — it's a **boot-time load that 404s on the restored deep-link before the session is resolvable**, after which the WebUI shows its standard 404 empty-state.

## Likely cause (cross-repo: desktop restore × WebUI boot)
v0.6.0 restores each tab's **deep-linked session URL** (`/c/<id>` / `/session/<id>`) instead of the root (#30/#34), and re-seeds the per-tab `hermes_profile` cookie. On boot the WebUI then calls `GET /api/session?session_id=<id>` (`static/sessions.js loadSession()`); a 404 there renders exactly this message:

> `static/sessions.js:985` → `"Session not available in web UI."`

Two plausible triggers for the boot-only 404 (need to confirm which):
1. **Profile/cookie timing.** The restored tab navigates to the deep link before the re-seeded `hermes_profile` cookie is in effect for that request, so `/api/session?session_id=<id>` is resolved under the **wrong profile** and 404s. A later in-app switch carries the correct profile, so it then resolves — matching "switch and switch back is fine."
2. **Server-readiness timing.** Behind the reverse proxy (Tailscale/Docker, cf. #28), the deep-link load races server/session-index readiness; the session isn't queryable yet on first hit but is moments later.

Hypothesis (1) is the stronger fit for "whatever session you land in when the app opens" + clean recovery on switch — it's the restored/active tab specifically, and a manual switch is what fixes it.

## What to check / fix shape
[...]

Issue #11, opened 2026-06-12
Portable Windows build: support user-level in-place update without the admin installer


nesquena-hermes:
## Summary
*"Auto-update makes me download the installer even though I'm using the portable app. ... or [fix it] since Cursor and VSCode default to user installs in localappdata."* (**Rod**)

Currently the portable Windows build disables self-update and points at Releases (by design -- the NSIS/MSI installer needs admin and would convert the portable copy into an installed app; see CHANGELOG v0.3.1, `updater.rs:27-47`). The ask: a user-level update path (install/update into `%LOCALAPPDATA%` without admin, like VSCode/Cursor) so portable users get in-place updates.

## Files
- `src-tauri/src/updater.rs` (portable branch L27-47)
- `src-tauri/tauri.conf.json` (NSIS install mode / per-user install)

Issue #8, opened 2026-06-12
Per-tab profile indicator (icon or color dot)


nesquena-hermes:
## Summary
*"The tabs are nice; it might be helpful to have some visual icon or dot color so I know what profile each tab represents."* (**b3nw**)

Pairs with the multi-tab profile isolation bug -- once a tab can be reliably on its own profile, show a small per-tab indicator (color dot / initial) for that profile.

## Files
- `src/shared/shell.js` / `shell.html` (tab rendering)
- `src-tauri/src/strip.rs` (`TabEntry`, `emit_tabs` payload -- include profile)

nesquena-hermes:
Reinforcing this from a fresh report — **b3nw** again: *"color code the tabs by profile or some visual identifier."* Same ask as this issue, so capturing here rather than filing a duplicate.

Grounding it in the current code so it's actionable:

- `TabEntry` (`src-tauri/src/state.rs`) carries only `label`, `title`, `attention` — **no profile field**. The strip snapshot (`strip::snapshot`) emits `tabs`, `active`, `mode`, `target` but nothing per-tab profile-wise.
- There's also no per-tab profile *concept* in the backend yet — a tab is just a webview pointing at `target_url`; the profile is whatever the WebUI client resolved inside it. So this is blocked on knowing each tab's profile in native, which ties to **#3** (sidebar shows wrong profile after tab switch).

Concrete path once profile-per-tab is wired:
1. Add `profile: Option<String>` to `TabEntry`; populate it (from the WebUI reporting its active profile, or from a per-tab connection once **#9** lands).
2. Include `profile` in the `emit_tabs` payload.
3. In `renderTabs` (`src/shell.html`), render a small color dot / initial. Derive a stable color by hashing the profile name (consistent across tabs/sessions) so "default" vs "work" vs "test" are visually distinct without manual config.

Suggest sequencing this after **#3** so the profile signal is reliable before we surface it visually.

PR #39, opened 2026-06-19
v0.6.2 — bug squash: fix Windows session-restore (real root cause), profile dot, tab rename, hide-bar hint

nesquena:
## What this is

A **v0.6.2 bug-squash** focused on fixes that are *genuinely* resolved this time — including two (`#30`/`#37`) that previous releases only *appeared* to fix. I evaluated every open issue independently, fixed the live bugs, then ran an **adversarial verification pass** on each fix (multiple angles + full tracing + live smoke-testing where possible). That pass found **four real holes in my own first-cut fixes**, which this PR then corrects. Details below so you can review the reasoning, not just the diff.

## ⚠️ The headline: why #30/#37 kept coming back

`#30` was marked fixed in v0.6.0 and `#37` in v0.6.1, yet both kept happening on Windows. Root cause (traced through wry + the WebUI server this round):

- The profile selector `hermes_profile` is a **session cookie** (no expiry — confirmed in `hermes-webui` `api/helpers.py:586`).
- **WebView2 discards session cookies** from a tab's on-disk data folder when the app restarts. So a restored tab — even reusing its data folder to keep you logged in — came back on the **default** profile.
- Its saved `/session/<id>` link is **profile-scoped**, so the server returns it not-found (`routes.py` `_session_visible_to_active_profile`) → "Session not available in web UI." → bounce to root.
- The earlier fixes were only ever validated on the **macOS forced-strip harness** (`HERMES_FORCE_STRIP=1`), which keeps the cookie in memory for the app's lifetime and so never reproduced the WebView2 behavior.

**Fix:** re-seed the saved profile cookie on restore, **pinned to the target host**. The host pin is the crux — `wry::WebView::set_cookie` takes no URL, so a domain-less cookie becomes `CreateCookie(.., domain="")` on WebView2 and is **silently never sent** (verified in `wry-0.55.1/src/webview2/mod.rs:1575`). That's why a naive re-seed compiles, runs, swallows no error, and does nothing on Windows. macOS native restore keeps the domain-less form WKWebView accepts, so it's left untouched.

## The adversarial pass — 4 holes found and fixed

I fanned out four independent skeptics (one per fix), each told to *refute* full resolution. They earned their keep:

| Hole | Severity | Fix in this PR |
|---|---|---|
| **#30/#37** re-seed used a **domain-less** cookie → dropped by WebView2 → fix was a no-op on Windows | Critical | Pin `set_domain(host)` on the re-seeded cookie |
| **#31** front-end `dot_profile \|\| tab.profile` fallback **resurrected a false dot** for the default profile (the WebUI persists `hermes_profile=default` on switch-to-default; also a renamed root) | High | Drop the cookie fallback (reporter is authoritative); filter the literal `"default"` from the seed |
| **#10** hint flag was **burned even if the toast was suppressed** (Focus Assist / notifications-off / dev build) → first-time hider permanently stuck | High | Retire the hint on the **un-hide** transition (proof the user learned Ctrl+Shift+B), not on hide |
| **#38** closing the tab being renamed left an **orphaned input** on a gone tab | Low | Release the rename + rebuild when the renamed tab leaves the snapshot |

(The #38 verifier also confirmed the headline "input vanishes while typing" is genuinely fixed and that the closing-a-tab-mid-rename path is near-unreachable in practice — any close action blurs the input first — but the guard is cheap insurance.)

## Issues fixed in this PR (5)

- **#30 + #37** — restored tab reopens on the right profile & real session instead of bouncing to home. (Root cause + fix above.)
- **#31** — per-tab profile dot correct on launch and never shown for the default profile. A page-side `/api/profile/active` reporter drives the dot authoritatively, instead of the cookie (which the WebUI sets only on an explicit switch).
- **#38** — renaming a tab no longer has the edit box vanish mid-keystroke. Emit `tabs-changed` only on a real change, and never rebuild the strip while a rename is open.
- **#10** — hiding the tab bar now shows an OS notification with the un-hide shortcut (the "⋯" button disappears with the bar, so first-timers had no affordance). (The "consolidate with the WebUI top bar" half of #10 remains an enhancement, not done here.)

## Every other open issue, triaged

- **#32** native notifications, **#26** dot-updates-on-switch, **#8** profile dot, **#7** rename, **#18** tab restore, **#19** drag-reorder, **#22** macOS one-tab drag — shipped/fixed in v0.4–v0.6.1; this PR builds on them (and #18 only truly works on Windows *with* the #30/#37 fix here).
- **#28** (502 behind a proxy on relaunch) — the per-tab boot retry added for #37 also helps the bounce case, but #28's specific "readiness probe accepts 5xx" concern was **not** re-verified this release; leaving open.
- **#5** (icon white fringe) — **deferred**: the interior near-white is shared with the logo glyph (sampled), so a recolor would damage the logo; needs corrected source art.
- **#9** (per-tab server connections), **#11** (portable user-level updater) — enhancements/feature requests, out of scope for a bug-squash.

## Verification

- ✅ `cargo fmt --check`, `cargo clippy` (clean), `cargo test` (14 pass, incl. `ssh_args_frozen`, `profile_cookie_matches_webui_attrs`).
- ✅ **Windows cross-compile** (`cargo xwin check --target x86_64-pc-windows-msvc`) — the WebView2 / notification / cookie paths compile.
- ✅ **Live smoke (macOS forced-strip)** of the #31 reporter chain against a fake `/api/profile/active`: non-default → dot, default → **no** dot, clean boot, no panic.

**Honest gap:** #30/#37's re-seed and #31's reporter are `cfg(not(target_os = "macos"))`, so their *runtime* behavior can't be exercised on this macOS box beyond the cross-compile, the live reporter smoke, and source-level tracing through wry + the WebUI server. The #30/#37 host-pin in particular wants a **real Windows restore test** (named profile + a `/session/<id>` tab, quit, reopen → lands on the session). Flagging it for that check before publish.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,52 @@
 # Changelog
 
+## [v0.6.2] — 2026-06-19
+
+A bug-squash release centered on the Windows session-restore failures that
+earlier releases only appeared to fix — they were validated on the macOS
+forced-strip harness, which doesn't reproduce WebView2's cookie behavior, so
+they shipped still broken on Windows. This release traces and fixes the actual
+root cause.
+
+### Fixed
+
+- **Windows: a restored tab now reopens on the right profile and its real
+  session, instead of bouncing to the home screen with "Session not available in
+  web UI."** (issues #30 and #37 — one root cause; #30 was marked fixed in v0.6.0
+  and #37 in v0.6.1, but both kept happening on Windows). The profile selector
+  (`hermes_profile`) is a *session* cookie, and WebView2 discards session cookies
+  from a tab's on-disk data folder when the app restarts. So a restored tab —
+  even though it reuses its data folder to keep you logged in — came back on the
+  *default* profile, and its saved `/session/<id>` link is profile-scoped, so the
+  server returned it as not-found and the tab bounced to the home screen. The app
+  now re-establishes the saved profile on restore by re-seeding that one cookie
+  before the tab loads, pinned to the server's host so WebView2 actually sends it
+  — a cookie set without an explicit host is silently dropped there, which is why
+  the earlier attempt looked right in testing yet did nothing on Windows. Your
+  login and drafts still ride along in the reused data folder; only the profile
+  selector is restored. (macOS was unaffected — its webview keeps the in-memory
+  cookie for the life of the app.)
+- **Windows/Linux: the per-tab profile dot is now correct on launch and never
+  shows for the default profile** (issue #31, follow-on to v0.6.0). The dot was
+  driven by the profile *cookie*, which the WebUI sets only on an explicit
+  switch — so a tab that simply *started* on a named profile showed no dot, while
+  a tab switched back to the default profile (or a renamed root profile) showed a
+  spurious one. Each tab's page now reports its actual active profile and that
+  drives the dot directly: it appears on the starting profile, and the default
+  profile correctly shows nothing.
+- **Windows/Linux: renaming a tab no longer has the edit box vanish as you type**
+  (issue #38). Double-clicking a tab opens an inline rename field, but the active
+  tab's title changes constantly while a chat streams, and each change rebuilt the
+  tab strip — destroying the open field mid-keystroke. The strip now rebuilds only
+  when something it actually shows has changed, and never while a rename is in
+  progress, so the field stays put until you press Enter or Escape.
+- **Windows: hiding the tab bar now tells you how to bring it back** (issue #10).
+  Hiding the bar (the "⋯" menu or Ctrl+Shift+B) also hides the "⋯" button — the
+  only on-screen control — leaving first-time users with no way to discover how to
+  restore it. Hiding it now shows an OS notification with the shortcut, and keeps
+  reminding you on each hide until you've successfully brought the bar back at
+  least once (so the hint isn't wasted if the notification was suppressed).
+
 ## [v0.6.1] — 2026-06-19
 
 A crash-and-bug-fix release.
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.6.0",
+  "version": "0.6.2",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.6.1"
+version = "0.6.2"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.6.1"
+version = "0.6.2"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -492,6 +492,35 @@ const ROUTE_REPORTER: &str = r##"
   })();
 "##;
 
+/// S15 — active-profile reporter (issue #31): report the tab's active profile
+/// NAME so the strip can show the profile dot on the STARTING profile, not just
+/// after a switch. The WebUI sets the `hermes_profile` cookie only on an
+/// explicit `/api/profile/switch` — never on boot — so the shell's cookie-based
+/// dot is blank for a tab sitting on its launch profile. The page always knows
+/// the name (`/api/profile/active` → `{name, is_default}`), so it reports it
+/// (empty for the default profile = no dot), debounced to fire only on change.
+const PROFILE_REPORTER: &str = r##"
+  (function () {
+    var last = null;
+    var report = function () {
+      try {
+        fetch('/api/profile/active', { credentials: 'same-origin' })
+          .then(function (r) { return r.ok ? r.json() : null; })
+          .then(function (p) {
+            if (!p) return;
+            var name = p.is_default ? '' : (p.name || '');
+            if (name !== last) { last = name; EMIT('profile', name); }
+          })
+          .catch(function () {});
+      } catch (e) {}
+    };
+    if (document.readyState === 'loading') document.addEventListener('DOMContentLoaded', report);
+    else report();
+    setInterval(report, 3000);
+    window.addEventListener('focus', report);
+  })();
+"##;
+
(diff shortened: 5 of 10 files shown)


nesquena:
# Deep review — approve, with one elevated verification concern

First: the self-adversarial pass and writeup are excellent — documenting the four holes you found in your own first cut (and *why* the prior #30/#37 "fixes" were no-ops) is exactly the reasoning that makes this reviewable. I independently verified the three load-bearing claims at the source and traced every change. No confirmed code bug — but I want to push harder on one thing the writeup is optimistic about, because it determines whether the headline fix actually works on the *default* Windows config.

## Verified at source (not taking the writeup's word)

- **WebView2 drops the domain-less cookie** — `wry cookie_into_win32` maps `cookie.domain() == None` → `CreateCookie(.., domain=HSTRING::new(), ..)`. An empty-domain cookie has no host to match, so it's created-but-never-sent. The crux holds, and `set_domain(host)` does produce a real Domain cookie. ✓
- **`/api/profile/active` exists** (`api/routes.py:7804`) — the #31 reporter has a real endpoint returning `{name, is_default}`. ✓
- **`hermes_profile` is a session cookie** (no Max-Age in `build_profile_cookie`). ✓
- gates green (fmt, clippy, **14 tests**), **`cargo xwin check` clean**.

## ⚠️ Elevated concern: does the host-pin work for `localhost` and IP targets, or only real hostnames?

The fix swaps a domain-less cookie for `set_domain(host)` — a **`Domain` cookie**. But Chromium (and therefore WebView2):
- **rejects `Domain` cookies for IP-literal hosts** (an IP can only have a host-only cookie — `Domain=192.168.x.x` / `Domain=127.0.0.1` is invalid and dropped), and
- treats **`Domain=localhost`** inconsistently (a dotless domain; historically rejected in some builds).

The comment calls `set_domain(host)` "host-only-equivalent for localhost," but a `Domain` cookie is *not* host-only, and that distinction is exactly what Chromium validates. Since the **default Direct target is `http://localhost:8787`** and self-hosters frequently point at a **LAN IP**, the headline #30/#37 fix may land correctly for hostname targets (Tailscale MagicDNS, real domains) yet **still drop the cookie for localhost/IP on Windows** — i.e., fix the case you can't easily test and miss the common one. This is the single thing to verify before the release notes claim #30/#37 are fixed: a real Windows restore of a named-profile `/session/<id>` tab against **each** of localhost, an IP, and a hostname. (macOS/Linux are fine — they keep the host-only form their engines accept.)

## Second-order concern: switch-after-restore may leave two `hermes_profile` cookies

The re-seed sets `Domain=host`; the WebUI sets `hermes_profile` **host-only** (no Domain) on an explicit switch. Per RFC 6265 those are *distinct* cookies (host-only-flag differs), so after a restore *then* an in-tab profile switch, the jar can hold both `Domain=host` (old value) and host-only (new value) → the request sends two `hermes_profile` values and the server picks by header order (unspecified). Worth checking: after a restored tab, switch profiles and confirm it switches cleanly (no revert/stale). If it's a problem, matching the WebUI's exact host-only form would be ideal — but that's the very form WebView2 drops, so it may need a URL-scoped cookie path (which `wry::set_cookie` lacks) or clearing the re-seeded cookie post-load.

## The rest — sound by trace

- **#31 dot** — driving the dot from the page-reported active-profile name (authoritative), dropping the `tab.profile` cookie fallback, and filtering the literal `"default"` correctly fixes both the missing-on-start and spurious-on-default cases. The dot is seeded from the creation profile so it degrades gracefully if the reporter's EMIT doesn't arrive. (Same EMIT-on-remote-WebView2 caveat as the #34 route reporter — but here the seed fallback makes it low-risk.)
- **#38 rename** — suppressing strip rebuilds while a rename input is open + emitting `tabs-changed` only on a real title change is the right two-part fix (the constant `document.title` churn was the root cause), and the orphan-input guard (release + rebuild if the renamed tab left the snapshot) is correct.
- **#10 hint** — retiring the one-time flag on the **un-hide** transition rather than the hide is exactly right: a suppressed toast (Focus Assist / notifications off / dev build) no longer permanently spends the hint. Windows-only, matching where the strip toggle runs.

## Recommendation

Approve. The fixes are well-reasoned, the prior-review items hold, and the code is correct as written — the open risks are WebView2 runtime behaviors that genuinely can't be exercised off-Windows. Just treat the localhost/IP host-pin question as a gating check before publishing #30/#37 as fixed (it's plausibly the difference between fixing the common case and not), and glance at switch-after-restore. Happy to help with a URL-scoped re-seed path if the Domain-cookie approach turns out to drop on localhost/IP.

nesquena:
Thanks — this is the right thing to push on, and I dug into both concerns. Short version: the host-pin isn't a new cookie mechanism (it's byte-identical to the new-tab seed we already ship), and switch-after-restore resolves cleanly. I'll let the **draft Windows build** be where the localhost/IP restore smoke actually runs, before Publish.

## The host-pin is the *same* cookie the new-tab path already creates

The key thing I'd undersold: the re-seeded cookie is **structurally identical** to the shipped new-tab seed.

- New tab (`add_tab`, the #3 per-tab-profile feature) seeds from `opener.cookies()`, and `cookie_from_win32` reads the host back as the domain — so it calls `set_cookie` → `CreateCookie(.., domain="localhost", ..)`.
- My restore re-seed sets `domain = url.host_str()`, and for the default target `http://localhost:8787` that's the string `"localhost"` — **the exact same value** `cookies()` reports. So restore now produces the identical cookie object the new-tab path produces; it's not a novel `Domain`-cookie path, it's bringing restore in line with the one we ship.

And that mechanism is **confirmed working on Windows + localhost**: #3 (per-tab profiles) was verified on the Windows app at v0.3.7 ("looks good" from the original reporter). So `CreateCookie(domain="localhost")` does produce a cookie that's sent on localhost — the default config is covered by precedent, not hope.

For an **IP target**: restore behaves *identically* to new-tab seeding (same `domain=host` call). So whatever Chromium does with `Domain=<IP>`, restore is **no worse than shipped** — if it ever drops there, new-tab profile inheritance is already affected and it's a separate, pre-existing item, not a regression introduced here.

## Switch-after-restore: resolves to the new value

Traced this end to end and it's benign:

- The server reads the cookie via Python `http.cookies.SimpleCookie().load(...)` (`api/helpers.py` / `auth.py:parse_cookie`) — a dict load where, on a duplicate name, **the last occurrence wins**.
- Chromium orders the `Cookie:` header by path length (both are `/`), then **creation-time ascending** — the re-seed (created at restore) precedes the server's switch cookie (created later), so the switch value is **last** → SimpleCookie keeps it. Switch takes effect.
- And if `CreateCookie` with a dot-less host yields a **host-only** cookie (matching the WebUI's exact form), the server's `Set-Cookie` simply *replaces* it — no duplicate at all.

Either way, a profile switch after a restore lands on the new profile.

## Gating check → on the draft

Since neither of us can exercise WebView2 off-Windows, the real restore smoke belongs on the draft artifact before Publish: a named-profile `/session/<id>` tab → quit → reopen, against **localhost**, an **IP**, and a **hostname**, plus a switch-after-restore. Proceeding to cut the draft so that test has a build to run on; not publishing until it's done.

Issue #47, opened 2026-06-22
FR: user-configurable custom color per agent profile (override the auto name-hash dot)


nesquena-hermes:
## Summary

The per-tab profile color dot (#8) currently derives its color by hashing the profile name (`profileColor(name)` → `hsl(hash % 360, 62%, 55%)` in `shell.html`). Users want to **set a custom color per profile** instead of relying on the auto-generated hash, so the dots are meaningful/distinguishable at a glance.

Requested by **@Lemz** in #desktop-apps (2026-06-21):

> maybe even set custom colours per agent profiles

## Relationship to existing dot issues (distinct from all of them)

- **#8** — added the per-tab profile dot (the auto name-hash color). This FR is about *overriding* that color.
- **#41** — "dot colors too visually similar": that's about improving the *automatic* palette spacing. A user-set custom color is an orthogonal, complementary feature (and would let a user manually fix any collision #41 doesn't).
- **#26/#31/#45** — dot *correctness* bugs (stale/wrong/collapsed color). This is a net-new customization feature, not a bug.

So this is genuinely new — none of the above lets a user pick the color.

## Fix shape

A persisted per-profile color override that `profileColor()` consults before falling back to the name-hash:

1. Store an optional `{profile_name: hex}` map in the desktop app's prefs (`src-tauri/src/prefs.rs`), or — if it should be portable across the web app too — server-side per-profile metadata (larger scope; desktop-local prefs is the smaller first cut).
2. `shell.html` `profileColor(name)` returns the override if present, else the current hash.
3. A small UI to set it — e.g. a color swatch in the tab context menu or the profile settings — writing the override back over the bridge.

Desktop-local prefs is the recommended first cut (smallest, ships the value immediately); a server-side per-profile color that the web app also honors is a possible follow-up.

## Scope

Desktop-wrapper feature (`src-tauri/src/prefs.rs` + `src/shell.html` + a small color-picker UI). Windows/Linux strip; would also inform the macOS-native indicator (#44).

## Requested by

@Lemz — Discord #desktop-apps, 2026-06-21. (Captured by the desktop bug monitor; filed by maintainer agent.)

nesquena:
[...]

Issue #46, opened 2026-06-22
FR: tab active/inactive (session-in-progress) indicator like the web app spinner


nesquena-hermes:
## Summary

The desktop strip tabs have no indication of whether a tab's session is actively working vs idle. In the web app you can tell a session is in progress from the small spinning circle; the desktop tab strip has no equivalent, so with multiple tabs you can't see which ones have a run in flight.

Reported by **@b3nw** in #desktop-apps (2026-06-19):

> when you're in the normal app, you can tell a session is in progress because of the small spinning circle, would be great if the tabs had an icon so you knew if they were active or inactive

## Why it fits the desktop shell

The strip already renders per-tab decorations: the attention dot (pending approval/clarify — `shell.html` `.attn`) and the per-profile color dot (`.pf`, #8/#31). An "active/working" indicator is the same class of per-tab status decoration, driven by whether that tab's session has an in-flight stream.

## Fix shape

The web page already knows its own busy state (the composer spinner / streaming state). Two ways to surface it on the native tab:

1. **Page-reported (preferred, consistent with the profile dot #31):** add a small reporter that posts the tab's busy/idle state over the existing bridge (same mechanism as `PROFILE_REPORTER` → `EMIT('profile', …)`), and render a spinner/active glyph on the tab in `shell.html` (a new `.active`/`.working` span alongside `.attn`/`.pf`). Debounce on change like the profile reporter does.
2. **Title-derived (fallback):** the active tab's `document.title` already gets a `●` prefix while streaming (`ui.js` ~7524). The shell could read that marker, but the bridge-reported state is cleaner and works for background tabs too.

Note the attention dot (waiting-for-you) and a working/active indicator are **distinct states** — one means "the agent needs you," the other means "the agent is running." Both can coexist on a tab.

## Scope

Desktop-wrapper feature (`src-tauri/src/bridge.rs` reporter + `src/shell.html` render + a small WebUI-side busy reporter if going the page-reported route). Windows/Linux strip; macOS native tabs would need the title-derived path (cf. #44 for the macOS-native dot equivalent).

## Reported by

@b3nw — Discord #desktop-apps, 2026-06-19. (Captured by the desktop bug monitor; filed by maintainer agent.)

nesquena:
[...]

Issue #45, opened 2026-06-21
All tab profile dots collapse to the same color after an update + restart (Windows/Linux strip)


nesquena-hermes:
## Summary

On the desktop app (Windows/Linux strip), **after applying an update and restarting, every tab's profile color dot collapses to the *same* color** — the per-tab profile identity that the dots normally encode is lost and all tabs render one identical dot.

Reported by **@b3nw** (Discord, 2026-06-21), with a screenshot:

> desktop specific bug, after applying an update and restarting the webui, all tabs dots turned the same color

## Evidence (pixel-measured, not eyeballed)

The forwarded screenshot shows a 7-tab strip. Sampling the center of each tab's dot:

```
tab 1  #93cf52      tab 5  #93cf52
tab 2  #93cf52      tab 6  #93cf52
tab 3  #93cf52      tab 7  #93cf52
tab 4  #93cf52
```

All seven are **byte-identical** `rgb(147, 207, 82)` = `#93cf52` (hue ≈ 89° at `hsl(_, 62%, 55%)`).

Two things follow from the measured value:

1. **This is NOT #41 ("colors too visually similar").** These dots are not merely close in hue — they are the *exact same* color. #41 is a palette-spacing problem between *distinct* profiles; this is a *collapse* to one profile.
2. **The shared color is a non-default profile's color.** `profileColor("default")` hashes to hue 345 (red `#d34569`); the empty/default name maps to red as well. Hue ≈89 (`#93cf52`) is the color of a single **named** profile. So all 7 tabs are reporting the **same profile name** to the strip — not 7 different ones, and not "default."

The active tab's title in the screenshot is **"Hermes WebUI Updater Lock Error"**, i.e. that tab is sitting on a WebUI update-error page — confirming the server was in an update/restart state when the collapse happened.

## How the dot is driven (so the collapse can be reasoned about)

The dot is authoritatively set, per tab, from the page's active-profile name:

- `PROFILE_REPORTER` (`src-tauri/src/bridge.rs:507`) fetches `/api/profile/active` on load, on navigation, and on a 3s/focus backstop, then `EMIT('profile', name)`.
- The bridge routes that to `set_tab_dot_profile` (`src-tauri/src/strip.rs:1045`) → `tab.dot_profile`.
- `shell.html:292` (`profileColor(tab.dot_profile)`) renders the dot.

[...]

Issue #44, opened 2026-06-19
macOS: no per-tab profile indicator — surface it Mac-native via the tab title (dot is Win/Linux strip only)


nesquena-hermes:
## Summary

On macOS there's **no per-tab profile indicator** — the profile color dot is Windows/Linux only. b3nw asked whether that's expected (it is, by design), but the underlying need — telling which profile a tab is on at a glance — applies on macOS too, and there's currently no Mac-native way to do it.

> *"on mac there are no tab dots, is that expected?"* — @b3nw, #webui, 2026-06-19, macOS v0.6.3

## Is it expected? Yes — and why

Confirmed by design. The profile dot lives in the **custom Windows/Linux tab strip**. The bridge reporter that feeds it (`PROFILE_REPORTER`, `src-tauri/src/bridge.rs:507`) is only injected on non-macOS — `bridge.rs:549-553` pushes it under `cfg!(not(target_os = "macos"))` with the comment *"Profile dot is strip-only (Windows/Linux); macOS native tabs have no dot."* macOS uses native NSWindow tabbing, which has no strip and nowhere to hang a colored dot. So a macOS user gets no profile indicator at all.

## The Mac-native fix shape (this issue)

Native macOS tabs show the **tab title**, sourced through `apply_reported_title` (`windows.rs:792`, fed by the `on_document_title_changed` hook at `:228`). That title is the natural carrier for a profile indicator on macOS — there's no strip, but there is a per-tab title. Options:

- Prefix the macOS tab title with a small, stable profile marker — a colored circle glyph (●) per profile, or the profile name in brackets — for non-default profiles. Reuse the same per-profile color mapping concept as the Win/Linux dot (and inherit the palette improvement tracked in #41 so colors stay distinguishable).
- The active-profile name is already available the same way the strip gets it (`/api/profile/active`); on macOS just route it into `apply_reported_title` composition instead of into a strip dot.

Keep it subtle and native — a Mac user should see a clean tab with a discreet profile cue, not a bolted-on widget.

## Scope

[...]

PR #48, opened 2026-06-22
v0.6.4 — profile dots (distinct / custom / reliable) + macOS tab discoverability

nesquena:
## What this is

A sweep of the open issues that were well-specced enough to fix confidently in one go, organized as two themes — **profile dots** (the most-reported area after v0.6.3) and **macOS tab parity** — plus a regression-locking test. I triaged every open issue (in-line + with parallel investigators); the verdicts are at the bottom.

### Profile dots
- **#41 — distinct dot colors.** v0.6.3 hashed the profile name straight onto the hue wheel with no minimum spacing, so distinct profiles often rendered near-identical (clustered pinks). Colors now come from a **curated, well-separated 12-color palette** indexed by name-hash — same profile → same color, distinct profiles → distinct colors. (`src/shell.html` `profileColor`.)
- **#45 — dots no longer collapse to one color after a server update/restart** (Windows/Linux). The dot was driven *solely* by the page-reported active profile (`/api/profile/active`), which — while the server restarts — can transiently return the same **process-global** profile for *every* tab, turning all dots one color (and the debounce kept them stuck). The dot now prefers each tab's **own per-tab `hermes_profile` cookie** (read from that tab's isolated jar — genuinely per-tab, so it can't collapse), falling back to the page report only for a tab still on its starting profile (no cookie yet). *Note:* the underlying transient is a WebUI-side resolution behavior (the issue's leading hypothesis, still unconfirmed); this is the desktop-side hardening that makes the dots immune to it. One line: `prof = tab.profile || tab.dot_profile`.
- **#47 — custom per-profile colors.** Right-click a tab → OS color picker → the color is persisted in app prefs (`#rrggbb`-validated, since it's injected into CSS) and consulted before the auto palette. Per-profile, so every tab on that profile recolors. Lets you override the auto color or disambiguate a palette collision from #41.

### macOS parity
- **#44 — per-tab profile indicator on macOS.** macOS uses native tabs (no color-dot strip), so a *separate* macOS-only profile reporter prefixes the native tab title with a non-default profile name (`work · …`); default stays unprefixed. Deliberately a separate `mac-profile` bridge message so the Win/Linux dot payload is untouched. **Verified live on a real macOS build** (`set_window_profile_name main-1 -> "work"`).
- **#42 — first-run "tabs exist" hint on macOS.** macOS tabs are easy to miss (no on-screen "+", native tab bar only appears with 2+ tabs). A one-time notification points at ⌘T, **retired when the user actually opens a tab** (not on show) so a dropped toast isn't permanently spent — the issue-#10 lesson. The File ▸ New Tab ⌘T item is the always-present fallback. **Verified live** (`#42 tabs-hint fired`).

### Test hardening
- **#28 — already fixed in v0.5.0** (`health::http_ready` rejects 502–504 so restore waits for a warming proxy instead of landing tabs on gateway-error pages). It was stale-OPEN; I added the missing **502–504 boundary unit test** to lock that load-bearing contract. **#28 is resolved → closeable.**

## Verification
- ✅ `cargo fmt`/`clippy` clean, **18 tests** pass (3 new `http_ready` boundary + 1 `is_hex_color`), **Windows cross-compile** clean.
- ✅ Live macOS smokes: #42 hint fires on connect; #44 sets the title prefix; forced-strip boots clean with the shell.html changes (no JS break); #41 palette distinctness checked (8 distinct for 10 profiles; #47 covers the rest).
- Honest gaps: #45's transient is partly a WebUI-side behavior — the desktop fix makes the dots immune regardless, but the root server behavior is the issue's (unconfirmed) hypothesis. The Win/Linux dot rendering is validated via build + forced-strip boot + the palette check (the dot reporter itself is `cfg(not macos)`, so its live wiring was proven in v0.6.3).

## Triaged but deferred (with reasons)
- **#46** (tab in-progress/working indicator) — feasible via the WebUI's `S.busy`, but whether an injected script can read that JS global is unverified; I won't ship a maybe-silent no-op. Design is ready for a follow-up (could use the server-side `is_streaming` for robustness).
- **#43** (macOS workspace→composer drop never lands) — WKWebView-internal; the fix is uncertain and not verifiable off a real drag. Needs a focused, macOS-driven session.
- **#11** (portable Windows user-level update) — Windows-installer-only; can't be tested from here. (Note: a `portable.zip` already ships as of v0.6.3.)
- **#9** (multiple/per-tab server connections) — a genuine feature, too large for this sweep.
- **#5** (icon white fringe) — needs corrected source art (the interior near-white is shared with the logo glyph).

## Already resolved by v0.6.2/v0.6.3 — closeable
**#37, #38, #31, #26** are fixed in shipped releases but still show OPEN (the merges didn't carry auto-close keywords). Safe to close. (#30, #33, #32, #8, #22, #18, #19, #7 already closed.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,41 @@
 # Changelog
 
+## [v0.6.4] — 2026-06-21
+
+A sweep focused on the per-tab profile dots — distinct, customizable, and
+reliable — plus macOS tab discoverability.
+
+### Added
+
+- **Custom per-profile dot colors** (issue #47). Right-click a tab and pick a
+  color for its profile; every tab on that profile takes it and the choice is
+  remembered. Handy to override the auto-assigned color or to disambiguate two
+  profiles that happen to share one.
+- **macOS: a per-tab profile indicator** (issue #44). macOS uses native tabs
+  (there's no Windows/Linux color-dot strip), so a tab on a non-default profile
+  now shows the profile name as a prefix in its native tab title (e.g.
+  "work · …"); the default profile stays unprefixed.
+- **macOS: a first-run hint that tabs exist** (issue #42). macOS tabs are easy
+  to miss — there's no on-screen "+", and the native tab bar only appears once
+  you have two tabs. The app now shows a one-time notification pointing at ⌘T
+  (also under File ▸ New Tab), and keeps the nudge available until you've opened
+  a tab at least once, so a missed notification isn't wasted.
+
+### Fixed
+
+- **Profile dot colors are now visually distinct** (issue #41). v0.6.3 mapped a
+  hash of the profile name straight onto the hue wheel with no minimum spacing,
+  so different profiles often rendered in near-identical hues (clustered pinks).
+  Colors now come from a curated, well-separated palette, so distinct profiles
+  are tellable apart at a glance (and #47 lets you override any clash).
+- **Profile dots no longer all collapse to one color after a server
+  update/restart** (issue #45, Windows/Linux). The dot was driven solely by the
+  page-reported active profile, which can transiently report the same
+  process-global profile for every tab while the server restarts — turning all
+  dots one color. Each tab's dot is now driven first by its own per-tab profile
+  cookie (genuinely per-tab, so it can't collapse), falling back to the page
+  report only for a tab still on its starting profile.
+
 ## [v0.6.3] — 2026-06-19
 
 A bug-squash release: the Windows session-restore failures fixed at the real
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.6.3",
+  "version": "0.6.4",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.6.3"
+version = "0.6.4"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.6.3"
+version = "0.6.4"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -537,6 +537,45 @@ const PROFILE_REPORTER: &str = r##"
   })();
 "##;
 
+/// macOS-only profile reporter (issue #44). macOS uses native tabs with no
+/// strip, so there's nowhere to hang the Win/Linux color dot — surface the
+/// active profile in the native tab TITLE instead. Mirrors `PROFILE_REPORTER`
+/// but reports the name ONLY for a non-default profile (empty = default = clear
+/// the prefix, so single-profile users' titles stay clean). A distinct
+/// `mac-profile` kind keeps the Win/Linux dot's `profile` payload untouched.
+const MACOS_PROFILE_REPORTER: &str = r##"
+  (function () {
+    var last = null;
+    var report = function () {
+      try {
+        fetch('/api/profile/active', { credentials: 'same-origin', cache: 'no-store' })
+          .then(function (r) { return r.ok ? r.json() : null; })
+          .then(function (p) {
+            if (!p) return;
+            var name = p.is_default ? '' : (p.name || '');
+            if (name !== last) { last = name; EMIT('mac-profile', name); }
+          })
+          .catch(function () {});
+      } catch (e) {}
+    };
+    var wrap = function (n) {
+      try {
+        var orig = history[n];
+        if (typeof orig === 'function') {
+          history[n] = function () { var r = orig.apply(this, arguments); report(); return r; };
+        }
+      } catch (e) {}
+    };
+    wrap('pushState'); wrap('replaceState');
+    window.addEventListener('popstate', report);
+    window.addEventListener('hashchange', report);
+    if (document.readyState === 'loading') document.addEventListener('DOMContentLoaded', report);
+    else report();
+    setInterval(report, 3000);
+    window.addEventListener('focus', report);
+  })();
+"##;
+
 /// Assemble the per-window initialization script.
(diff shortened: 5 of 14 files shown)


nesquena:
# Deep review — approve, with one auth-mode issue worth deciding on before relying on the dots

Clean sweep, and most of it is solid. I traced every change and verified the load-bearing bits at the source (including the WebUI). One confirmed degradation in **auth mode** is the only thing I'd flag as more than cosmetic-minor; everything else checks out. No crash, no data loss.

## ⚠️ #45 colors/labels the dot by the raw `hermes_profile` cookie — which is a signed token in auth mode

#45 switches the dot source to `prof = tab.profile || tab.dot_profile`, where `tab.profile` is the raw `hermes_profile` cookie value read from the tab's jar. That's the right instinct for the collapse bug (the per-tab cookie can't go process-global the way `/api/profile/active` can). But I checked `build_profile_cookie` in `hermes-webui`:

- **No-auth mode:** `value = name` → the cookie is the plain profile name → everything works (clean color + tooltip). ✓
- **Auth mode:** `value = sign_profile_cookie_value(name, …)` → the cookie is a **signed, session-bound token**, not the name.

So with auth enabled (the shared/remote-server setup this app's SSH tunneling targets), `prof` becomes that token, and:
1. **Tooltip** shows the opaque token — `· profile: <base64-ish blob>` — instead of `· profile: work`.
2. **Color inconsistency:** a tab with the cookie colors by `hash(token)`, while a tab still on its starting profile colors by `hash(name)` (via `dot_profile`) — so two tabs on the *same* profile can show *different* colors. That's the opposite of #41's "same profile → same color" goal.
3. **#47 custom colors** are keyed by `prof` = the token, which is session-bound — so a custom color won't reattach after re-login.

No-auth users see none of this. But for auth users it noticeably degrades the very feature this PR is polishing. Suggested fix: drive the dot's **color and tooltip from the profile name** (`dot_profile`, the `/api/profile/active` name) and use the per-tab cookie only as the *signal that the profile differs* — or detect auth mode and fall back to `dot_profile` for display. (The collapse #45 fixes is real, so keep the per-tab preference; just don't use the raw cookie value as the display identity.) Worth a quick auth-mode check before the release notes lean on the dots.

## Minor: default-profile dot is inconsistent

`/api/profile/active` reports `is_default` (confirmed `routes.py`), and the reporter sends `''` for it — so a tab **still on the default profile** (no `hermes_profile` cookie) shows **no dot**, while a tab **switched to default** (cookie set) shows one. v0.6.3's changelog said "dot on every tab incl default," so this is at least inconsistent — worth deciding whether default should be dotted at all, then making it uniform.

## #47 custom colors — safe (checked the injection surface)

The validation is strict (`is_hex_color` = exactly `#` + 6 hex, with the injection-shape rejections unit-tested), and just as importantly the color only ever reaches `pf.style.background = …` (a DOM `.style` setter, which parses a single property value and can't break out or execute JS) and `input[type=color].value` (accepts only `#rrggbb`). There's no `cssText`/`innerHTML`/`<style>` text sink, so even an unvalidated value couldn't inject. Solid. (Minor defense-in-depth note: `prefs::profile_colors` returns the stored map unvalidated on read, so a hand-edited `prefs.json` bypasses `is_hex_color` — fully mitigated by the safe sink, but re-validating on read would close it entirely.)

## The rest — sound by trace

- **#44 macOS indicator** — the `name · base` prefix sits outside `display_title`'s truncation (never cut), uses a distinct `mac-profile` bridge kind so the Win/Linux dot payload is untouched, and is per-window (correct for macOS one-window-per-tab). Clean.
- **#42 tabs hint** — fires once per launch (atomic guard) and retires on the user's first tab-open, not on show, so a suppressed toast isn't spent — exactly the #10 lesson, correctly applied. (Retiring only on `as_tab=true`, not new-window, is the right call.)
- **#41 palette** — curated 12-color palette indexed by name-hash is a real improvement over the unspaced hue-hash; same-profile→same-color holds (modulo the auth-mode identity issue above).
- **#28 test** — extracting `ready_from_status` and testing the 502–504 boundary (plus 401/404/405/500/501 = ready) correctly locks invariant #2's readiness refinement.
- gates: `fmt`/`clippy` clean, **18 tests** pass, **`cargo xwin check` clean**.

## Recommendation

Approve. The only thing I'd resolve before relying on the dots for auth-mode users is the cookie-token-as-identity issue in #45 (color/label by the profile name instead) — it's confirmed for auth mode, not speculative, but it's cosmetic and no-auth is unaffected, so it's a "decide and fix or consciously accept," not a blocker. The default-dot inconsistency is a smaller product call. Everything else is good to ship. Happy to wire the name-based dot identity if you want it.

nesquena:
Thanks — the auth-mode catch was a real one. Pushed `33a544f` addressing all three points.

## ✅ #45 auth-mode token-as-identity — fixed (kept the collapse-resilience)

You're right: under auth `hermes_profile` is `sign_profile_cookie_value` → `name.<64-hex-hmac>`, so coloring/labeling by the raw cookie value was wrong in auth mode. But I didn't want to give up the per-tab collapse-resilience (the per-tab cookie is the only signal immune to `/api/profile/active` going process-global). The key realization from the WebUI source: the **name is recoverable** — `sign_profile_cookie_value` returns `f"{profile_name}.{sig}"`, and profile names can't contain `.` (`_PROFILE_ID_RE = ^[a-z0-9][a-z0-9_-]{0,63}$`). So a trailing `.<64 hex>` is unambiguously the signature.

New `profileName()` strips it: `cookieValue.replace(/\.[0-9a-f]{64}$/, "")`, and the dot is now `profileName(tab.profile) || tab.dot_profile`. Result in **both** modes:
- Tooltip shows the real name (`· profile: work`), not a blob.
- Color is `hash(name)`, so two tabs on the same profile match (switched or starting).
- #47 overrides key off the name → survive re-login.
- Still per-tab → still immune to the #45 collapse.

(Verified the strip: `"work"`→`"work"`, `"work.<64hex>"`→`"work"`, `"my-team_2"`→unchanged, default handled.)

## ✅ #47 read-validation — added

`profile_colors()` now re-validates each entry with `is_hex_color` on read, so a hand-edited `prefs.json` can't route a non-`#rrggbb` value into `style.background`. (The sink was already safe as you noted; this closes it at the source too.)

## Default-profile dot — clarifying (it's two different reporters)

I think this conflated the two reporters:
- **Win/Linux dot** (`PROFILE_REPORTER`): sends `p.name` unconditionally (incl. `"default"`) since v0.6.3 — so a tab on default **does** show a dot, whether or not it has a cookie. Consistent with v0.6.3's "dot on every tab incl default." (Confirmed `var name = p.name || ''`.)
- **macOS title** (`MACOS_PROFILE_REPORTER`, new in #44): sends `''` for `is_default`, so default tabs get **no title prefix** — deliberate, so single-profile macOS users don't get `default · ` clobbering every tab title.

So Win/Linux is uniform (default dotted); macOS deliberately doesn't prefix default. Happy to revisit the macOS choice, but prefixing every default tab seemed worse. No code change for this one.

Gates green after the fixes: fmt/clippy, **18 tests**, forced-strip boots clean. Moving to release.

Issue #53, opened 2026-06-22
Desktop: model-selection dropdown shows a stale/wrong model list on session visit; Providers refresh corrects it (desktop-only)


nesquena-hermes:
Desktop: model-selection dropdown shows a stale/wrong model list on session visit; Providers refresh corrects it (desktop-only)
## Symptom
In the desktop app, opening an existing session shows an unexpected/stale list of models in the model-selection dropdown. Manually navigating to Providers and triggering a refresh repopulates the dropdown with the expected model list. The issue is reported as desktop-only: the same server viewed in a normal (non-desktop) web browser shows the correct model list.

## Platform
Desktop app (wrapper). Reported on the desktop build; not observed in a normal desktop web browser pointed at the same server.

## Suspected area
Repo ambiguous. Either (a) the desktop shell's per-tab webview serving a stale cached model list (hermes-desktop-rust), or (b) a WebUI-side model/provider cache (e.g. SESSION_AGENT_CACHE / provider model fallback list in api/config.py) that only surfaces under the desktop wrapper's persistent webview session (nesquena/hermes-webui). Distinct from webui #4706 (fresh OpenRouter-only install shows a small default list with no search) and #4684 (Studio model selection session-scoping).

## Repro
1. Open an existing session in the desktop app.
2. Observe the model-selection dropdown lists an unexpected/stale set of models.
3. Go to Providers and hit refresh.
4. Re-open the dropdown — the expected model list now appears.
5. (Compare) Same session in a normal desktop web browser shows the correct list without the refresh.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Resolved upstream — closing. This was a server-side model-cache freshness issue (the desktop sends `Cache-Control: no-store` and the service worker skips `/api/*`, so the webview wasn't caching it). Fixed in **hermes-webui** (#4756, shipped v0.51.615): the model catalog now refreshes on session visit when the profile-scoped cache is stale, without a manual Providers refresh. Once your server is on v0.51.615+, the dropdown shows the current models on visit. No desktop change required.

Issue #52, opened 2026-06-22
Desktop new tab: default workspace not auto-selected, and workspace can't be changed mid-turn


nesquena-hermes:
Desktop new tab: default workspace not auto-selected, and workspace can't be changed mid-turn
## Symptom
Opening a new desktop tab does not auto-select the expected workspace. The wrong workspace is only noticed after the first prompt is sent, and the workspace cannot be changed until the active turn finishes.
## Platform
Desktop app (wrapper).
## Suspected area
Two facets: (a) new-tab default-workspace auto-selection — ambiguous between desktop tab/profile mechanics (cf. closed desktop-rust #3, profile switch intermittently dropping a profile's default workspace) and WebUI default-workspace selection on a new session; (b) workspace selector locked during an active streaming turn — a WebUI surface behavior that would reproduce in a normal browser. Repo is ambiguous (desktop-rust vs webui).
## Repro
1. Open a new tab in the desktop app.
2. Observe the workspace is not the expected one.
3. Send a prompt; notice the workspace is wrong.
4. Attempt to change the workspace — it is locked until the turn completes.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Resolved upstream — closing. This was WebUI app logic (the desktop just loads the server root URL; no desktop-side workspace logic exists). Fixed in **hermes-webui** (#4755, shipped v0.51.612): a fresh session now honors the profile's default workspace. Once your server is on v0.51.612+, a new desktop tab lands on the default workspace.

Note: the *mid-turn* workspace-switch block (while a turn is streaming) was kept intentionally as a tested contract upstream — that part is working-as-designed, not a bug. If you want mid-turn switching to become possible, that's a separate WebUI feature request.

Issue #51, opened 2026-06-22
Desktop: notification re-appears in every tab on app re-open — add a universal/cross-tab dismiss-or-action


nesquena-hermes:
Desktop: notification re-appears in every tab on app re-open — add a universal/cross-tab dismiss-or-action
## Symptom
When the desktop app is re-opened with multiple tabs open, the same notification/popup re-appears independently in every single tab instead of once. The user wants a universal (cross-tab) way to click-to-dismiss or action the notification, rather than handling the same notification separately in each tab.

## Platform
Windows/Linux tab-strip mode, desktop app (wrapper). Reported on Windows.

## Suspected area
Desktop shell (cross-tab coordination). Each strip tab is its own webview, so an outstanding web-rendered notification re-displays per tab on relaunch; a universal dismiss/action would need shell-level coordination across tabs. Note: the notification itself is rendered on the WebUI surface, so attribution is mildly ambiguous between the desktop shell and WebUI.

## Repro
1. Open the desktop app with multiple tabs.
2. Have an outstanding notification/popup present.
3. Quit and re-open the app.
4. Observe the same notification appears in every tab; dismissing it in one tab does not clear the others.

## Related (distinct, not duplicates)
- #32 (bridge web notifications to native OS notifications)
- #14 / #17 (badge a tab with a pending approval/clarify popup)
- #50 (notification toggle discoverability)

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Half of this ships in **v0.7.0**: the OS-notification side is now deduped natively — N tabs observing the same event produce ONE notification instead of one per tab (30s window, at the single bridge choke point). What remains is the **in-page popup** re-appearing per tab and a cross-tab dismiss — that's WebUI-rendered UI, so the clean fix is webui-side (broadcast dismissals over its own channel or localStorage). Keeping open, rescoped to the in-page half.

Issue #50, opened 2026-06-22
Desktop notifications toggle not discoverable — lives in WebUI settings, expected in desktop Preferences


nesquena-hermes:
Desktop notifications toggle not discoverable — lives in WebUI settings, expected in desktop Preferences
## Symptom
A desktop user could not get notifications working until they enabled browser notifications in the general WebUI settings. They expected a desktop-app-specific notification toggle in the desktop Preferences (where the server connection is configured), and assumed it would live there as a desktop-app setting. Suggests surfacing or mirroring the notification-enable toggle (or a pointer to it) in the desktop Preferences.
## Platform
Windows, desktop app (wrapper)
## Suspected area
Discoverability — notification enablement is currently on the WebUI settings surface; users expect it in the native desktop Preferences. Mild repo ambiguity: the toggle today lives on the WebUI surface, so this could be a desktop-shell Preferences addition or a WebUI settings change. Related to but distinct from the existing native-notification-bridging work.
## Repro
1. Open the desktop app Preferences (where the connection is configured).
2. Look for a notification enable/disable option.
3. Observe there is none; notifications only work after enabling browser notifications in the general WebUI settings.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Shipped in **v0.6.5**. The desktop notifications toggle now sits under its own NOTIFICATIONS heading in Preferences and notes that notifications must also be enabled in Hermes WebUI Settings → "Browser notifications" (the second gate users had to discover separately).

PR #54, opened 2026-06-23
v0.6.5 — tab-strip + notification fixes + per-tab working indicator (#19, #46, #49, #50)

nesquena:
## What this is

A v0.6.5 bug-fix release. I re-reviewed every open issue, closed the one now verifiably complete (**#45**), and fixed the three that are cleanly, verifiably fixable on the desktop side. The rest of the open set is triaged at the bottom (most are WebUI-side or would-regress — not responsibly includable in a desktop-only release).

## Fixed

- **#19 — Windows/Linux: drag-to-reorder tabs now actually works.** The reorder handlers shipped in v0.4.0 but were dead on Windows/Linux: the tab-strip webview never called `.disable_drag_drop_handler()`, so wry's native OS drag-drop handler intercepted the gesture and the strip page never saw the HTML5 `dragstart`/`drop` events — the *exact* root cause the #27 fix handled for content webviews, just never applied to the strip shell webview. One-line fix at the strip webview builder (`strip.rs`). (macOS reorders via its native tab bar, a different path.)
- **#49 — the per-profile color picker has clear controls.** v0.6.4's right-click picker just `.click()`d a hidden `<input type="color">`, so users got the bare OS color dialog with no obvious apply/close (only "click away" dismissed it). Replaced with a small in-strip popover: one-click palette swatches, a custom-color input, and explicit **Reset** (→ auto color) / **Close** (+ Escape / click-away). Backend unchanged — still persists via `set_profile_color` (which clears on a non-hex value).
- **#50 — the desktop notifications toggle is discoverable, and points at the second gate.** The toggle already existed in Preferences but was buried and gave no hint that notifications *also* require the WebUI's own "Browser notifications" setting (the switch a user had to find before anything fired). It now sits under its own **NOTIFICATIONS** heading with that pointer.

## Closed (verified complete)
- **#45** (dots collapse after restart) — resolved by v0.6.4's per-tab-cookie dot (structurally can't collapse) **and** the WebUI root-cause fix #4586 (closed). Commented + closed.

## Verification
- `cargo fmt`/`clippy` clean, **18 tests** pass, **Windows cross-compile** clean.
- Forced-strip boots clean with the shell.html changes; `shell.html` script syntax-checked (`node --check`). #19's true bug is WebView2-specific, so a Windows artifact is the authoritative check, but the fix mirrors the proven #27 pattern.

## Triaged but NOT included (with reasons)
- **WebUI-side (out of scope for a desktop release):** **#52** (default workspace / mid-turn switch — confirmed WebUI app logic), **#53** (stale model dropdown — confirmed the WebUI's profile-scoped server-side model cache; `/api/models` is already `no-store` in the webview), and most of **#51** (the in-page approval cards re-surface per session = server state).
- **Would regress / needs WebUI coordination:** **#32** (background-tab notifications) — the only desktop lever is forcing `document.hidden` on hidden tabs, which would trip the WebUI's *close-SSE-on-hidden* logic and break the "hidden tabs stay alive" guarantee. Needs a desktop-aware notify path in the WebUI.
- **Needs a wry fork / page-side change:** **#43** (macOS WKWebView drop) — wry 0.55 can't split "don't swallow OS file-drops" from "deliver in-page HTML5 drops" on WKWebView; not fixable here without a fork or a hermes-webui MIME change (and unverifiable without manual drag testing).
- **Needs source art:** **#5** (icon corner fringe) — the source is a pre-rounded badge with a baked light rim; the clean fix is a full-bleed square master (none in-repo). No safe pixel-only transform (the rim is entangled with the gradient + the chrome logo).
- **Windows-unverifiable, reported case already covered:** **#37** (restore "Session not available") — v0.6.3's host-pinned re-seed + 2.5s retry already cover the reported hostname/proxy case; the residual `Domain=localhost`/IP concern is medium-risk and needs a real Windows test, not a blind change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,30 @@
 # Changelog
 
+## [v0.6.5] — 2026-06-22
+
+A bug-fix release for the Windows/Linux tab strip and notification discoverability.
+
+### Fixed
+
+- **Windows/Linux: dragging a tab now reorders it** (issue #19). The reorder
+  handlers shipped in v0.4.0 but never worked: the tab-strip webview didn't
+  disable wry's native OS drag-drop handler, so it intercepted the gesture and
+  the strip page never received the HTML5 `dragstart`/`drop` events (the same
+  root cause the #27 fix addressed for content webviews). The strip webview now
+  disables the native handler, so drag-to-reorder works. (macOS already reorders
+  via its native tab bar.)
+- **Windows/Linux: the per-profile color picker now has clear controls** (issue
+  #49, follow-up to v0.6.4's #47). Right-clicking a tab previously opened the
+  bare OS color dialog with no obvious way to apply or close it. It now opens a
+  small in-strip popover with one-click palette swatches, a custom-color input,
+  and explicit **Reset** (back to the auto color) and **Close** buttons (also
+  Escape / click-away). The chosen color still persists per profile.
+- **Notifications: the desktop toggle is now easier to find, with a pointer to
+  the second switch** (issue #50). The desktop Preferences notification toggle
+  now sits under its own "NOTIFICATIONS" heading and notes that notifications
+  must *also* be enabled in Hermes WebUI Settings → "Browser notifications" —
+  the WebUI-side gate a user had to discover separately before anything fired.
+
 ## [v0.6.4] — 2026-06-21
 
 A sweep focused on the per-tab profile dots — distinct, customizable, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,18 @@
 
 ## [v0.6.5] — 2026-06-22
 
-A bug-fix release for the Windows/Linux tab strip and notification discoverability.
+A bug-fix release for the Windows/Linux tab strip and notification discoverability,
+plus a per-tab "working" indicator.
+
+### Added
+
+- **Windows/Linux: tabs now show a "working" spinner while their session is
+  streaming** (issue #46). A small spinner appears on a tab whose session has a
+  run in flight (mirroring the web app's in-progress spinner), so with several
+  tabs open you can see at a glance which background tabs are busy. Driven by the
+  page's own busy state — the same signal as the in-app spinner — and clears the
+  moment the run finishes (the tab's profile dot / attention badge takes over
+  again).
 
 ### Fixed
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.6.4",
+  "version": "0.6.5",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.6.4"
+version = "0.6.5"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.6.4"
+version = "0.6.5"
 description = "Hermes WebUI Desktop — cross-platform shell for hermes-webui"
 edition = "2021"
 license = "MIT"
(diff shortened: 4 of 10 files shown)


nesquena:
# Deep review — #19 and #50 good to ship; #49 needs rework (the popover can't render in the 38px strip)

Two of the three fixes are clean. The third (#49) has a confirmed functional defect: the new color popover is drawn into a webview that's too short to show it, so it ends up clipped and unusable. Walkthrough below.

## ✅ #19 — drag-to-reorder: correct

Adding `.disable_drag_drop_handler()` to the strip webview is exactly right and mirrors the proven #27 fix for content webviews — wry's native OS drag-drop handler was swallowing the HTML5 `dragstart`/`drop` the reorder code (shipped v0.4.0, index math verified back in the #24 review) depends on. The strip hosts only the tab bar (no file-drop target), and window-drag via `data-tauri-drag-region` is a separate mechanism, so disabling the OS handler loses nothing. The reorder gesture is horizontal within the 38px strip, so no height issue. Verification gap is the one the PR already flags: #19's true bug is WebView2-specific, so a real Windows drag is the authoritative check.

## ✅ #50 — notifications discoverability: fine

Pure Preferences markup — a NOTIFICATIONS heading and a hint pointing at the WebUI's "Browser notifications" second gate. Valid HTML, no behavior change. Good call surfacing the second gate.

## 🔴 #49 — the in-strip color popover is clipped to 38px and won't display

v0.6.4's #47 picker was bare but **functional**: `input[type=color].click()` opens the OS's own color dialog — a separate OS window, not confined to the webview. #49 replaces that with an **in-page DOM popover** (~90px tall: header + 6-wide swatch grid + custom/Reset/Close row). The problem: that popover renders inside the **strip webview, which is `STRIP_HEIGHT = 38px`**:

- `build_strip_window` adds the shell.html webview at `(0,0)` sized `(width, 38)`; the content tab webview is a *separate* native surface below it, and shell.html's `html, body { overflow: hidden }` clips to the 38px viewport.
- A `position: fixed` element is relative to that 38px viewport. The popover's own clamp computes `top = Math.min(y, window.innerHeight - h - 6) = Math.min(y, 38 - 90 - 6) = -58`, so it's pushed ~58px above the strip — only a sliver lands in the visible 38px band, and the swatches / Reset / Close never show.

(Rendered mock of this is in my message above — the swatches and buttons fall below the dashed 38px boundary, onto the content webview's surface, where the strip's DOM can't paint.) So the headline UX improvement is, in practice, worse than what it replaces — the old OS dialog at least appeared. The PR's checks wouldn't surface this: "forced-strip boots clean" and `node --check` don't open the popover or test its visibility (and forced-strip on macOS has the same 38px strip).

**Fix options** — the controls need a surface that isn't height-confined to the strip:
1. **Native submenu off the ⋯ menu** — reuse the existing `ContextMenu`/`popup` path (which already works and isn't clipped): a "Profile color ▸" submenu with swatch items, "Custom…", and "Reset". Most robust cross-platform, no new window plumbing.
2. **A small dedicated borderless webview window** positioned over the strip (taller than 38px) hosting the popover HTML.
3. **Keep the OS color dialog** (functional today) and add a separate **Reset** affordance elsewhere (e.g., a ⋯-menu item) — less elegant but cross-platform and shippable now.

Growing the strip webview while the popover is open is tempting but fights constraint #1 (resizing GTK child webviews crashes on Linux), so I'd avoid it.

## Recommendation

Ship #19 and #50 — both are correct and valuable. Rework #49 onto a non-strip-confined surface before it goes out (option 1 reuses the proven menu path and is probably the least work). Happy to wire the ⋯-menu submenu version if that's the direction. Everything else — gates green (18 tests, clippy, Windows cross-compile clean), and nice to see v0.6.4 went out "auth-correct."

nesquena:
**Update — added #46 since the initial review.** Pushed `a3e73ef`: a per-tab "working" spinner for the Windows/Linux strip (the session-in-progress indicator).

- **What:** a tab whose session is actively streaming shows a small spinner (mirrors the web app's in-progress spinner), so background tabs with a run in flight are visible at a glance.
- **How:** a strip-only injected `BUSY_REPORTER` polls the WebUI's `S.busy` global → `EMIT('busy')` → `set_tab_busy` → `TabEntry.busy` → `.working` glyph in `shell.html`. Highest-priority leading decoration (busy > attention > profile dot), yields when the run clears.
- **The one risk I de-risked:** whether an injected script can read the WebUI's `S.busy` (a top-level `const` in the classic, deferred `ui.js`). **Verified live** on forced-strip against a faithful fake page (`const S={busy:true}` via a deferred `ui.js`): the reporter read it and the busy chain fired (`tab-1-1 -> true`).
- **Verification:** fmt/clippy clean, 18 tests, Windows cross-compile clean, `shell.html` JS syntax-checked.

This is the only change since your review — #19/#49/#50 are unchanged. Worth a quick second look at the #46 diff (`bridge.rs` BUSY_REPORTER + arm, `strip.rs` set_tab_busy, `state.rs` field, `shell.html` glyph) before we ship.

Issue #57, opened 2026-06-24
Desktop: workspace-tab file download saves silently — no destination picker or completion notification


nesquena-hermes:
Desktop: workspace-tab file download saves silently — no destination picker or completion notification
## Symptom
Downloading a file from the workspace tab in the desktop app saves silently: there is no save dialog to pick the destination folder and no notification/confirmation that the download completed. The file just lands somewhere without user feedback.
## Platform
Desktop app (wrapper) — workspace tab download.
## Suspected area
Desktop shell download handling. The Mac-native app already intercepts WKWebView downloads with a native save panel (swift-mac #66), so the Rust shell likely lacks the equivalent native download handling (no Save-As dialog / no completion notice). Distinct from the workspace→composer drag-drop work (#27). Some scope ambiguity: treat as a bug (downloads should surface a destination picker) and/or an enhancement (add a native save dialog + completion notification).
## Repro
1. Open the desktop app and a workspace tab.
2. Download a file from the workspace tree.
3. Observe: no destination picker and no completion notification; the file is saved silently.

---
_Filed from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Fixed in **v0.6.6**: workspace-tree downloads on macOS/Linux now open a native **Save panel** (choose destination + filename), then **notify on completion and reveal the file** in Finder/Files — exactly the two asks here (destination picker + completion feedback). Windows already used WebView2's own Save dialog. From the v0.6.6 changelog: *"Workspace-file downloads show a Save dialog and reveal the file (issue #57, macOS/Linux)."*

Closing as shipped — please reopen if any download path still saves silently on ≥ v0.6.6.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #56, opened 2026-06-24
Desktop: WebUI upgrade triggered in one tab still prompts in the other tabs (no cross-tab coordination)


nesquena-hermes:
Desktop: WebUI upgrade triggered in one tab still prompts in the other tabs (no cross-tab coordination)
## Symptom
In the Windows/Linux desktop app with multiple tabs open, triggering a WebUI upgrade from one tab does not suppress the upgrade prompt in the other tabs — the remaining tabs still offer/continue to show the upgrade prompt after one tab has already initiated the upgrade.
## Platform
Windows/Linux desktop app (wrapper) — multi-tab strip.
## Suspected area
Cross-tab coordination of the upgrade prompt. AMBIGUOUS between the desktop shell (each strip tab is its own webview, so the prompt should be coalesced/dismissed across tabs by the shell) and the WebUI surface (the upgrade prompt itself is web-rendered and may re-show per tab). Related in shape to the per-tab notification/approval re-appearance (desktop-rust #51 / webui #4754) but this is the upgrade prompt specifically.
## Repro
1. Open the desktop app with two or more tabs.
2. In one tab, trigger the WebUI upgrade.
3. Observe the other tabs still offer the upgrade prompt instead of coordinating/dismissing it.

---
_Filed from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena-hermes:
Related symptom on **v0.6.9**, reported by **b3nw** (Discord):

> after update to underlying web app, all tabs show a "must hard refresh" message, requiring extra clicks, either auto-refresh them (if not active?) or enable a "refresh all" option.

Same cross-tab-coordination surface as this issue, but the *post*-upgrade side: once the underlying WebUI has updated, **every** tab independently shows a "must hard refresh" banner and each one has to be dismissed/refreshed by hand.

Requested behavior (either or both):
- **Auto-refresh** tabs that aren't currently active once the new WebUI is live (the active tab could prompt, background tabs just reload).
- A single **"Refresh all"** action so it's one click instead of N.

Filing note: this is arguably the natural completion of #56 — coordinate the upgrade prompt across tabs (#56 proper) *and* coordinate the post-upgrade hard-refresh across tabs (this comment). If the fix diverges, split this into its own issue.

nesquena:
[...]

Issue #55, opened 2026-06-24
Tab loses its per-profile color dot while the session-in-progress spinner is active (Win/Linux strip)


nesquena-hermes:
## Symptom
On the Windows/Linux tab strip, the per-tab session-in-progress spinner now works, but while a tab's spinner is active (a session is running) that tab loses its per-profile color dot — the color indicator drops/reverts while spinning and only returns once the spinner stops.

## Platform
Windows/Linux desktop app (wrapper) — tab strip.

## Suspected area
Desktop shell, tab strip rendering. Looks like an interaction/regression between the v0.6.5 per-tab session-in-progress spinner (the shipped activity indicator) and the per-tab profile color dot (a Windows/Linux strip feature): the active spinner state appears to override or replace the color-dot rendering on the same tab.

## Repro
1. Open the Windows/Linux desktop app with at least one profile tab.
2. Confirm the tab shows its profile color dot while idle.
3. Start a session in that tab so the in-progress spinner appears.
4. Observe: while the spinner is active, the tab's profile color dot loses its color; it returns when the spinner stops.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Fixed in **v0.6.6** and confirmed in the shipped code: the per-tab working spinner's ring is now tinted with the tab's profile color (`src/shell.html`, indicator block), so a busy tab keeps showing which profile it's on for the whole run — the dot's identity job is preserved even while the spinner has the slot. From the v0.6.6 changelog: *"A working tab keeps its profile color (issue #55)... the spinner is now tinted with the profile's color."*

Closing as shipped — @rod if the tint isn't landing for you on ≥ v0.6.6 (e.g. with a custom #47 color), please reopen with a screenshot.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #49, opened 2026-06-22
Color picker popup: no obvious way to confirm/apply or close (Win/Linux tab strip)


nesquena-hermes:
Color picker popup: no obvious way to confirm/apply or close (Win/Linux tab strip)
## Symptom
The per-profile color picker popup (Windows/Linux tab strip, shipped in v0.6.4) has no obvious affordance to confirm/apply the chosen color or to close the popup. Dismissal only happens by clicking away from the popup, which is not discoverable. The color picker itself works and the chosen color persists across restarts.
## Platform
Windows, desktop app (wrapper)
## Suspected area
Desktop shell — the Win/Linux tab strip color picker UI (follow-up polish on the v0.6.4 picker; distinct from the original FR and the release issue).
## Repro
1. Open the per-profile color picker for a tab on the Windows/Linux strip.
2. Choose a color.
3. Observe there is no explicit confirm/apply or close control; the popup only dismisses by clicking elsewhere.

---
_Filed automatically from a Discord community report. Edit, re-route, or close if mis-scoped._

nesquena:
Shipped in **v0.6.5**. Right-clicking a tab now opens an in-strip color popover with one-click palette swatches, a custom-color input, and explicit Reset / Close (plus Escape / click-away) — replacing the bare native dialog that had no obvious confirm/close.

PR #58, opened 2026-06-24
v0.6.6 — background-tab notifications, macOS Show-All-Tabs, download Save dialog, profile-colored spinner (#32, #42, #57, #55)

nesquena:
## What this is

A larger batch: I re-reviewed every open issue, closed the resolved ones, checked the WebUI repo for landed blockers that unlock desktop work, and packed in four fixes/features.

## The WebUI unlock (the headline)
Three of the blockers I filed last round have shipped in `hermes-webui`:
- **#4753 closed** → unlocked **#32** (I implement the desktop companion below).
- **#4755 closed** (v0.51.612) → resolves **#52** upstream.
- **#4756 closed** (v0.51.615) → resolves **#53** upstream.

(#4754 and #4757 are still open, so #51 and #43 stay blocked.)

## Fixed / Added (4)

- **#32 — background tabs now fire OS notifications.** A non-visible tab never fired its completion/approval notification, because the WebUI gates on `document.hidden` and a hidden tab's webview still reports visible. The WebUI shipped exactly the contract I proposed (`window.__hermesSetBackgrounded(bool)`, verified to feed the notification gate **only**, not SSE-close). The shell now calls it on tab switch (`strip::select_tab`/`add_tab`) and on app focus change (a new `WindowEvent::Focused` arm, both platforms) — so a background tab notifies *and* keeps streaming. No-ops harmlessly on a WebUI older than v0.51.612.
- **#42 — macOS Window ▸ Show All Tabs (⌃⌘T).** Summons the native tab bar with one window open (it otherwise only appears with 2+ tabs) — the Mac-native menu half of the discoverability work begun in v0.6.4. Implemented via `macos::toggle_tab_bar` → `NSWindow.toggleTabBar:`, **GCD-deferred** through `dispatch_main_async` (invariant #12, the freeze class). **Verified live**: a new `HERMES_TEST_TOGGLE_TABBAR` heartbeat hook drove the toggle and all 8 main-thread heartbeats continued afterward — no deadlock.
- **#57 — workspace downloads get a Save dialog + reveal.** Downloads used to save silently into ~/Downloads with no destination choice or feedback. Now a native Save panel (tauri-plugin-dialog, off the main thread) lets you pick location + name, then a completion notification fires and the file is revealed in Finder/Files (tauri-plugin-opener). mac/Linux (Windows already uses WebView2's own dialog).
- **#55 — a working tab keeps its profile color.** v0.6.5's per-tab "working" spinner hid the profile dot while streaming; the spinner is now tinted with the profile's color, so you still see which profile a busy tab is on.

## Closed (resolved) this round
- **#19, #46, #49, #50** — shipped in v0.6.5 (auto-close hadn't fired).
- **#52, #53** — resolved upstream (hermes-webui #4755 / #4756); no desktop change. Commented + closed.

## Verification
- `cargo fmt`/`clippy` clean, **18 tests**, **Windows cross-compile** clean.
- `shell.html` JS syntax-checked; forced-strip boots clean (shell.html + the #32 evals run without crashing); **#42 toggle deadlock-guarded live** on a real macOS build.
- Manual checks worth doing before publish: a real **download** on macOS (the native Save panel can't be auto-driven here), and **#32** end-to-end against a v0.51.612+ server (background tab → completion → OS notification).

## Still open / not in this release
- **#56** (cross-tab upgrade prompt) — **WebUI-side**: the `#updateBanner` is gated by per-webview `sessionStorage`, and the strip's tabs are storage-isolated by design, so coordination belongs in the WebUI. I can file it there (like #4754/#4757) if you want.
- **#51 / #43** — still blocked on WebUI #4754 / #4757 (open).
- **#5** (icon — needs a full-bleed square master), **#37** (Windows restore — needs a Windows smoke), **#9/#10/#11** (enhancements/larger).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,38 @@
 # Changelog
 
+## [v0.6.6] — 2026-06-24
+
+A batch of bug fixes and features: background-tab notifications, a Mac-native
+tab-bar menu, a download save dialog, and a profile-colored working spinner.
+
+### Added
+
+- **macOS: Window ▸ Show All Tabs (⌃⌘T) summons the native tab bar** even with
+  one window open (issue #42). macOS only shows the tab bar once you have two
+  tabs, so tabs were easy to miss — this is the Mac-native menu affordance that
+  completes the discoverability work begun in v0.6.4 (the first-run ⌘T hint).
+
+### Fixed
+
+- **Background tabs now fire OS notifications** (issue #32, with hermes-webui
+  #4753). A tab that wasn't the visible one never fired its completion/approval
+  notification: the WebUI gates notifications on `document.hidden`, and a hidden
+  tab's webview still reports visible. The shell now tells the page when its tab
+  is backgrounded — on tab switch, and when the whole app loses focus — and the
+  WebUI honors that for notifications only, so a background tab notifies *and*
+  keeps streaming. (Needs a hermes-webui new enough to include the notification
+  hook, v0.51.612+; on older servers it harmlessly no-ops as before.)
+- **A working tab keeps its profile color** (issue #55). v0.6.5's per-tab
+  "working" spinner hid the profile color dot while a run streamed; the spinner
+  is now tinted with the profile's color, so a busy tab still shows which
+  profile it's on.
+- **Workspace-file downloads show a Save dialog and reveal the file** (issue #57,
+  macOS/Linux). Downloads from the workspace tree used to save silently into
+  ~/Downloads with no destination choice or clear feedback. The app now opens a
+  native Save panel (choose the location + name), then notifies on completion
+  and reveals the file in Finder/Files. (Windows already used WebView2's own
+  Save dialog.)
+
 ## [v0.6.5] — 2026-06-22
 
 A bug-fix release for the Windows/Linux tab strip and notification discoverability,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ tab-bar menu, a download save dialog, and a profile-colored working spinner.
 
 ### Added
 
-- **macOS: Window ▸ Show All Tabs (⌃⌘T) summons the native tab bar** even with
+- **macOS: Window ▸ Show Tab Bar (⌃⌘T) summons the native tab bar** even with
   one window open (issue #42). macOS only shows the tab bar once you have two
   tabs, so tabs were easy to miss — this is the Mac-native menu affordance that
   completes the discoverability work begun in v0.6.4 (the first-run ⌘T hint).
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hermes-webui-desktop",
-  "version": "0.6.5",
+  "version": "0.6.6",
   "private": true,
   "scripts": {
     "tauri": "tauri"
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1706,7 +1706,7 @@ checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
 
 [[package]]
 name = "hermes-webui-desktop"
-version = "0.6.5"
+version = "0.6.6"
 dependencies = [
  "arboard",
  "base64 0.22.1",
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "hermes-webui-desktop"
-version = "0.6.5"
+version = "0.6.6"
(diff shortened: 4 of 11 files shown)


nesquena:
# Deep review — approve; I pushed a fix for the red CI

Solid batch. I traced all four changes, verified the load-bearing #32 contract in the WebUI source, and found one thing worth acting on immediately: **CI was failing**, so I pushed a fix (details first).

## 🔧 Fixed the red CI (`f39097c`)

`gh pr checks` showed the Test workflow failing — specifically the Clippy step (`cargo clippy -- -D warnings`, which CI enforces). The two `wv.eval(&format!(…))` calls added for #32 (`main.rs:549`, `strip.rs:752`) trip `clippy::needless_borrow` — `Webview::eval` takes `impl Into<String>`, so `format!()` needs no `&`. I dropped both, matching the existing `eval(format!(…))` pattern in `paste.rs`/`windows.rs`; behavior is identical. Clippy/`fmt`/18 tests are green locally afterward (with CI's exact `clippy -- -D warnings` command), and CI is re-running on the push. (Your local clippy must be a slightly older stable than CI's — worth aligning so "clippy clean" matches what CI sees.)

## #32 — background-tab notifications: correct, and verified at the source

I checked the WebUI side since the whole "notifies **and** keeps streaming" claim rests on it: `__hermesSetBackgrounded` sets `_desktopBackgroundedForNotifications`, which feeds **only** `_isBackgroundedForBrowserNotification()` (the notification gate — comment: "keep that signal notification-only"); it does **not** touch the SSE-close path. So a backgrounded tab notifies while its stream stays alive — the "hidden tabs stay alive" guarantee holds. ✓

The shell-side state machine is correct across every transition I traced:
- **new tab** → previous active tab hidden + `backgrounded(true)`; the new visible tab defaults to foreground (`document.hidden=false`), so not setting it explicitly is fine.
- **select_tab** → previous `backgrounded(true)`, new `backgrounded(false)`.
- **WindowEvent::Focused** → drives only the *active* tab (strip: `set_active_tab_backgrounded`; macOS: the window's own webview). Hidden tabs were already flagged by `select_tab` and correctly stay backgrounded regardless of app focus.

The `eval` is fire-and-forget (no nested runloop), so it's safe from the `Focused` tao callout. No-ops on a pre-v0.51.612 server.

## #57 — download Save dialog: correct

`blocking_save_file` is run **off the main thread** (the #33 lesson — a modal dialog on the UI thread would hang), with clean cancel (`None` → log + return) and write-error handling, then notify + `reveal_item_in_dir`. The 30MB-capped bytes clone into the worker is bounded and fine. mac/Linux only (Windows uses WebView2's dialog). ✓

## #42 — Show All Tabs: works, one label nit

`toggle_tab_bar` → `NSWindow.toggleTabBar(None)`, **GCD-deferred** via `dispatch_main_async` — the same invariant-#12-safe pattern as `add_tabbed_window`, and you live-verified it (8 heartbeats post-toggle). Functionally it does summon the bar for a single window. One nit: the menu item is labeled **"Show All Tabs" / ⌃⌘T**, but it calls `toggleTabBar:` — in macOS, "Show All Tabs" is the tab *overview* (`toggleTabOverview:`), whereas `toggleTabBar:` is "Show/Hide **Tab Bar**." Also it's a toggle with a static "Show" label (says "Show" even when the next click hides). Consider relabeling to "Show Tab Bar" (closer to the action) — purely cosmetic.

## #55 — profile-colored spinner: fine

Tinting the spinner border via `profileColor(prof)` reuses the safe `.style` sink and keeps the profile visible while busy. Guards on `prof` so the default profile keeps the plain spinner. ✓

## Recommendation

Approve once CI goes green on `f39097c` (re-running now). The fixes are correct, the #32 contract is verified end-to-end, and the only follow-ups are the cosmetic #42 label and the two manual checks you flagged (a real macOS download through the Save panel, and #32 against a v0.51.612+ server). Nice batch — and good call confirming the WebUI blockers landed before building the desktop companions.

nesquena:
Thanks — and thanks for catching + fixing the red CI (`f39097c`). Confirmed the `eval(&format!(…))` → `eval(format!(…))` fix is in; I re-ran with CI's exact `cargo clippy -- -D warnings` (exit 0) locally to match what CI sees going forward.

Addressed the #42 label nit in `ed797a3`: the item now reads **"Show Tab Bar"** (and the id is `show_tab_bar`) to match what it actually calls (`NSWindow.toggleTabBar:` = Show/Hide Tab Bar), rather than implying the tab *overview*. Cosmetic, no behavior change.

The two manual checks you flagged are noted for post-merge verification (a real macOS download through the Save panel, and #32 end-to-end against a v0.51.612+ server) — both are environment-dependent and can't be auto-driven here, consistent with how we've handled native-dialog / live-server behaviors in prior releases.

Proceeding to merge + release once CI is green on `ed797a3`.

PR #68, opened 2026-06-30
Linux: fix the large empty gap under the tab strip (content webview placed with physical bounds) (#67)

### LINKED ISSUE #67: Linux: large empty gap above the WebUI under the tab strip — content webview pushed down by a wrong scale factor
## Symptom

On Linux, the WebUI renders **below a large empty/black band** under the 38px tab strip, squashing the app into the lower portion of the window. Reported with a screenshot on a Linux desktop build (192.168.x VM display).

## Root cause

The Win/Linux tab strip (`strip.rs`) positions each tab's content webview as an X11 **child window** at `y = 38px` (the strip height), passing **logical** coordinates.

wry's WebKitGTK backend (`webkitgtk/mod.rs`, `create_container_x11_window` / `set_bounds`) re-derives its **own** scale factor for child windows from the X screen's reported physical size:

```
scale_factor_from_x11 = screen_width_px * 25.4 / screen_width_mm / 96
```

On VMs, Xvfb, remote/forwarded X, and any display where the X server misreports its physical mm size, this value is bogus (e.g. ~4.5). wry then does `Logical(38).to_physical(4.5)` ≈ **171 physical px**, so the content webview starts ~171px down instead of 38px — the gap. (Confirmed numerically against the `dpi` crate; ~171px matches the observed offset in the CI `linux-smoke` screenshot.)

macOS (WKWebView) and Windows (WebView2) place child webviews with the window's real scale, so they're unaffected.

## Why CI didn't catch it

`linux-smoke` runs headless under Xvfb (no WM) and only asserts "connected" + uploads a screenshot for human review — the gap is visible in that screenshot but isn't asserted on.

## Fix

Pass **physical** coordinates (computed from the window's real GTK scale factor) for the strip + content child webviews on Linux. wry passes `Physical` through with an identity cast, so geometry lands exactly under the strip regardless of the X server's DPI, and stays correct under genuine HiDPI.

nesquena-hermes:
Closes #67.

## The bug

On Linux the WebUI renders **below a large empty/black band** under the 38px tab strip, squashing the app into the lower part of the window (reported via Discord screenshot on a Linux desktop build). 

## Root cause

The Win/Linux strip (`strip.rs`) positions each tab's content webview as an X11 **child window** at `y = 38px`, passing **logical** coordinates. wry's WebKitGTK backend re-derives its **own** scale factor for child windows from the X screen's reported physical size:

```
scale_factor_from_x11 = screen_px × 25.4 ÷ screen_mm ÷ 96
```

On VMs / Xvfb / remote-forwarded X / any display that misreports its mm size, that value is bogus. wry then computes `Logical(38).to_physical(~4.5)` ≈ **171px**, pushing the content webview ~171px down — the gap. macOS (WKWebView) and Windows (WebView2) use the window's real scale, so they're unaffected.

I confirmed the mechanism end-to-end against the pinned dep sources:
- `wry 0.55.1` `webkitgtk/mod.rs`: `create_container_x11_window` / `set_bounds` apply `scale_factor_from_x11`; `Position::Physical` is an **identity cast**, `Position::Logical` is a **multiply**.
- `tauri-runtime-wry 2.11.2` `RectWrapper` + `with_bounds`: the `Position`/`Size` enum variant is preserved straight through to wry (no logical normalization).
- `tao 0.35.3`: Linux `scale_factor()` is the **GTK integer** widget scale (the correct render scale), not the bogus X11 float.

And reproduced the arithmetic numerically against the `dpi` crate: logical-38 → **171px** (matches the observed gap), physical-38 → **38px** (fixed), HiDPI scale-2 → 76px (correct).

## The fix

Place the strip + content child webviews with **physical** coordinates (computed from the window's real GTK scale factor) on Linux. wry passes `Physical` bounds through with an identity cast, so geometry lands exactly under the strip regardless of the X server's DPI, and stays correct under genuine HiDPI (scale 2). macOS/Windows keep logical bounds. Centralized in two small helpers (`child_position` / `child_size`) used by `content_bounds` and the strip-webview placement.

## Why CI didn't catch it

`linux-smoke` runs headless under Xvfb and only asserts "connected" + uploads a screenshot for human review — the gap is visible in that artifact (content text starts at ~y170) but wasn't asserted on. This PR's CI run will rebuild and produce a fresh smoke screenshot for visual confirmation that the page now sits directly under the strip.

## Testing

- Verified the fix arithmetic against the real `dpi` crate (logical→171 vs physical→38, plus HiDPI scale-2→76).
- `cargo fmt --check` clean.
- Full Linux compile + visual confirmation rely on CI here (the GTK/WebKit dev libs aren't available in my local env).

> Version bump (3 files) intentionally left to the releaser per the repo's release convention.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,26 @@
 # Changelog
 
+## [v0.6.8] — 2026-06-30
+
+### Fixed
+
+- **Linux: the WebUI no longer renders below a large empty gap under the tab
+  strip** (issue #67). On many Linux setups (VMs, Xvfb, remote/forwarded X, and
+  any display where the X server misreports its physical size) the content
+  webview was
+  pushed hundreds of pixels down the window, leaving a tall black band beneath
+  the 38px tab strip and squashing the app into the lower portion of the window.
+  Root cause: the shell positions each tab's webview as an X11 child window, and
+  wry's WebKitGTK backend re-derives its *own* scale factor for child windows
+  from the X screen's reported DPI (`screen_px × 25.4 ÷ screen_mm ÷ 96`). When
+  that DPI is wrong, the logical `38px` strip offset we passed got multiplied by
+  the bad factor (≈4.5× on the affected display → a ~171px gap). The strip and
+  content webviews are now placed with **physical** coordinates computed from
+  the window's real GTK scale factor, which wry passes through unchanged — so
+  the geometry lands exactly under the tab strip regardless of the X server's
+  DPI, and stays correct on genuine HiDPI displays. macOS and Windows are
+  unaffected (their webview backends already use the window's real scale).
+
 ## [v0.6.6] — 2026-06-24
 
 A batch of bug fixes and features: background-tab notifications, a Mac-native
--- a/src-tauri/src/strip.rs
+++ b/src-tauri/src/strip.rs
@@ -16,10 +16,50 @@ use std::sync::atomic::Ordering;
 use tauri::webview::{Color, Cookie, WebviewBuilder};
 use tauri::window::WindowBuilder;
 use tauri::{
-    AppHandle, Emitter, LogicalPosition, LogicalSize, Manager, Webview, WebviewUrl, Window, Wry,
+    AppHandle, Emitter, LogicalPosition, LogicalSize, Manager, PhysicalPosition, PhysicalSize,
+    Position, Size, Webview, WebviewUrl, Window, Wry,
 };
 
 pub const STRIP_HEIGHT: f64 = 38.0;
+
+/// Convert a logical child-webview position to the unit the platform's wry
+/// backend wants for `add_child` / `set_position`.
+///
+/// macOS (WKWebView) and Windows (WebView2) place child webviews using the
+/// window's OWN scale factor, so logical coordinates land correctly. WebKitGTK
+/// (Linux) is the exception: wry creates the child as an X11 sub-window and
+/// re-derives its OWN scale factor from the X screen's reported physical size
+/// (`scale_factor_from_x11` = screen_px * 25.4 / screen_mm / 96). On VMs, Xvfb,
+/// remote/forwarded X and many real setups that value is bogus (X reports a
+/// nonsense mm size), so wry multiplies our logical 38px strip offset by a
+/// wrong factor and the content webview is shoved hundreds of pixels down —
+/// the big empty gap under the tab strip, never caught by the headless smoke
+/// test (which runs against a fake X screen). Passing PHYSICAL coordinates
+/// (computed with tao's real GTK scale factor) makes wry's internal
+/// `to_physical()` an identity cast, so the geometry lands exactly where
+/// intended regardless of the X server's DPI lie — and stays correct under
+/// genuine HiDPI (scale 2).
+fn child_position(x: f64, y: f64, scale: f64) -> Position {
+    if cfg!(target_os = "linux") {
+        PhysicalPosition::new((x * scale).round() as i32, (y * scale).round() as i32).into()
+    } else {
+        LogicalPosition::new(x, y).into()
+    }
+}
+
+/// Companion to [`child_position`] for sizes — see that doc for why Linux needs
+/// physical units while macOS/Windows take logical.
+fn child_size(w: f64, h: f64, scale: f64) -> Size {
+    if cfg!(target_os = "linux") {
+        PhysicalSize::new(
+            (w * scale).round().max(0.0) as u32,
+            (h * scale).round().max(0.0) as u32,
+        )
+        .into()
+    } else {
+        LogicalSize::new(w, h).into()
+    }
+}
 
 /// Strip mode is the tab implementation everywhere except macOS.
 pub fn enabled() -> bool {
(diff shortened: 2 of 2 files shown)


nesquena-hermes:
## Verification + an honest note on CI coverage

I confirmed the root cause empirically (not just by reading source):

- Started `Xvfb -screen 0 1400x900x24 -dpi 1200` and probed it: the X server reports `1400x900 pixels (30x19 millimeters)`.
- wry's `scale_factor_from_x11 = 1400 × 25.4 ÷ 30 ÷ 96 ≈ 12.3`.
- A 38px strip offset × ~11–12 ≈ **418–456px** — which matches the ~450px gap in the reported screenshot, where the WebUI is shoved down to roughly the lower third of the window.
- Cross-checked numerically against the real `dpi` crate: `Logical(38).to_physical(12.3) → 469px` (the bug) vs `Physical(38).cast() → 38px` (the fix), and HiDPI `scale 2 → 76px` (correct).
- Verified end-to-end through the pinned deps that the `Position`/`Size` enum variant is preserved from `add_child` → `tauri-runtime-wry` `RectWrapper`/`with_bounds` → wry, so passing `Physical` reaches wry's X11 path as `Physical` (identity cast), bypassing the bad multiply.

**CI limitation (being upfront):** the `linux-smoke` job runs under `xvfb-run`, which normalizes the screen DPI back to sane and — critically — GTK clamps DPI to an *integer* widget scale, so under Xvfb `tao`'s scale and wry's X11 scale stay aligned and the gap does **not** reproduce there. I ran a throwaway high-DPI CI experiment that screenshotted buggy-vs-fixed; they came out pixel-identical (only the tab label differs), confirming the headless harness can't exercise this path. The bug requires a display where wry's X11 float diverges from GTK's integer scale (VMs / certain real displays that misreport physical mm), which is the reporting environment.

The change is provably a **no-op anywhere the layout is already correct** (scale 1 → 38 logical == 38 physical; macOS/Windows keep logical), so it can't regress those. @AvidFuturist — best verified by running a build of this branch on the affected machine; happy to cut a tagged build if you want to test the artifact directly.

nesquena (approved):
## Review: ✅ Approve

Reviewed the diff against the full `strip.rs` and traced the root-cause claim through the **pinned dependency sources** (not just the PR description). The fix is correct, complete, and a provable no-op where layout already works.

### Independent verification of the mechanism (at the source)

The fix rests on three claims; I confirmed each in the vendored crates that `Cargo.lock` actually pins:

1. **wry 0.55.1** `webkitgtk/mod.rs` `create_container_x11_window` (the `add_child` creation path) does:
   ```rust
   .map(|b| b.position.to_physical::<f64>(scale_factor))   // scale_factor = scale_factor_from_x11
   ```
   `dpi`'s `to_physical` is a **multiply for `Logical`** (the bug: `38 × bogus`) but an **identity for `Physical`** (the fix). And `scale_factor_from_x11` = `screen_px × 25.4 / screen_mm / 96` — a float off the X server's mm size, exactly as described.

2. **tao 0.35.3** `linux/window.rs`: `scale_factor: Rc<AtomicI32>` updated via `connect_scale_factor_notify` from GTK's `window.scale_factor()` — the **integer** widget scale (1/2/…). So `win.scale_factor()` is the *correct* render scale and genuinely **diverges** from wry's bogus X11 float. (If they were the same value, passing physical would reproduce the bug — they are not.)

3. **tauri-runtime-wry 2.11.2**: `RectWrapper` copies `position`/`size` **verbatim**, and the create path (`webview_attributes.bounds → with_bounds`) passes the original `Rect` straight to wry. The `to_logical` calls there only feed `auto_resize` rate math — and `add_child` doesn't enable `auto_resize`, so they're inert here. The `Physical` variant reaches wry intact.

So: `child_position` multiplies logical by the real GTK integer scale → wry identity-casts it → geometry lands at `38 × gtk_scale` physical, under the strip, regardless of the X DPI lie. Verified arithmetic: scale 1 → 38, scale 2 → 76 (HiDPI correct), bug display → 38 instead of ~171–469.

### Completeness

The only places Linux child-webview geometry is actually *applied* are the two **creation-time `add_child` calls** — strip webview (`build_strip_window`) and content webview (`add_tab_with` via `content_bounds`) — and both now go through the helpers. Every `set_position`/`set_size` site (`select_tab`, `layout`) is already Linux-gated (those crash on GTK children, invariant #1), so they correctly stay logical on mac/Win and no-op on Linux. Not a partial fix.

### No-regression

`child_position`/`child_size` return `Logical` unchanged on macOS/Windows, and at GTK scale 1 `38 logical == 38 physical` — so the change is a literal no-op everywhere layout is already correct. It cannot regress mac, Windows, or a correctly-reporting Linux display. `test` (fmt + clippy `-D warnings` + cargo test) is green.

### Honest note on confirmation

This is verified by source + arithmetic and is provably non-regressing — but the headless `linux-smoke` can't *reproduce* the bug (Xvfb + GTK's integer clamp keep tao and wry aligned, so the gap never appears there; the screenshot only confirms "no regression at scale 1"). True on-device confirmation comes from running a build on an affected display — which the v0.6.8 release artifact provides.

### Minor / non-blocking

- The helper doc says "≈4.5× → 171px" while the PR comment says "~12.3 → ~450px" — two different illustrative displays; harmless.
- CHANGELOG jumps v0.6.6 → v0.6.8 (v0.6.7's icon/version bump never got an entry). Pre-existing gap, not this PR's job.
- Future caveat: tauri-runtime-wry *does* `to_logical`-normalize on the `SetPosition`/`SetSize` message paths, so if anyone ever un-gates `set_position`/`set_size` on Linux they'd reintroduce this — the existing "never set bounds on GTK children" comments already steer away from that.

Approving and moving this forward as **v0.6.8**.

PR #73, opened 2026-07-11
Probes: trust OS-installed CAs — Test Connection works behind enterprise/internal PKI (#60)

### LINKED ISSUE #60: Test Connection fails for servers with enterprise/internal CA certificates
## Description

The "Test Connection" button in Preferences reports "Unreachable" for servers that are reachable and fully functional in a browser. This happens when the server uses a TLS certificate signed by an enterprise or internal CA (e.g., Caddy Local Authority, corporate PKI, self-signed CA imported into the OS trust store).

## Steps to Reproduce

1. Have a `hermes-webui` server behind a reverse proxy (e.g., Caddy) that serves TLS with a certificate from an internal/self-signed CA that is trusted by the OS but **not** in the Mozilla root store.
2. Open Preferences → enter the server URL (e.g., `https://ycj.ai.hylink.net.cn`)
3. Click "Test Connection"
4. **Actual:** "✗ Unreachable"
5. **Expected:** "✓ Connected" (the server returns HTTP 302 or 200, as confirmed by browser and `curl --ssl-no-revoke`)

## Root Cause

`health.rs` uses `ureq` with default features, which means `rustls` + `webpki-roots` (Mozilla root CA store) for TLS verification:

```rust
// health.rs — http_reachable (current)
let agent = ureq::AgentBuilder::new()
    .timeout_connect(timeout)
    .timeout(timeout)
    .build();
```

`webpki-roots` only trusts CAs in the Mozilla root program. Enterprise/internal CAs (e.g., Caddy Local Authority, corporate PKI roots) are **not** in that store, even when they are trusted by the operating system and browsers. A TLS handshake fails with `UnknownIssuer`, which `ureq` surfaces as a transport error → `http_reachable` returns `false`.

### Why browsers work but `ureq` does not

| Component | Trust Store | Internal CA Trusted? |
|-----------|-------------|---------------------|
| Chrome / Edge (Windows) | Windows Cert Store (SCHANNEL) | ✅ (CA imported by IT) |
| Safari (macOS) | macOS Keychain (Security.framework) | ✅ (CA imported by IT) |
| `hermes-swift-mac` (`URLSession`) | Platform trust store | ✅ |
| `hermes-desktop-rust` (`ureq` + `rustls`) | Mozilla root store (`webpki-roots`) | ❌ |

## Real-World Scenario

This is our production setup:
- A `hermes-webui` instance behind Caddy reverse proxy
- Caddy uses its own internal CA (`Caddy Local Authority - 2026 ECC Root`) to issue certificates for `*.ai.hylink.net.cn`
- The CA root certificate is deployed to all corporate Windows machines via Group Policy
- Browsers work fine, `curl --ssl-no-revoke` works, but `hermes-desktop-rust` "Test Connection" fails

Server response (confirmed reachable):
```
HTTP/1.1 302 Found
Location: login?next=/
Server: HermesWebUI Python/3.13.5
Via: 1.1 Caddy
```

## Suggested Fixes

Several options, ordered by preference:

### Option A: Use platform native certificate verification (recommended)

Enable `ureq`'s `native-certs` feature to load the OS trust store alongside the Mozilla roots:

```toml
# Cargo.toml
ureq = { version = "2", features = ["native-certs"] }
```

This uses `rustls-native-certs` which loads root CAs from:
- **Windows:** Windows certificate store via SCHANNEL
- **macOS:** Keychain via Security.framework  
- **Linux:** `/etc/ssl/certs` or similar

No code changes needed in `health.rs` — the `AgentBuilder` picks up the expanded trust store automatically. This matches what browsers do: trust both the Mozilla roots AND the platform-specific enterprise roots.

### Option B: Match the Swift app's behavior

The Swift app uses `URLSession` which inherently uses the platform trust store. For parity, use `native-tls` as the ureq TLS backend on all platforms:

```toml
ureq = { version = "2", default-features = false, features = ["native-tls"] }
```

This delegates TLS entirely to the platform (SCHANNEL on Windows, Security.framework on macOS, OpenSSL on Linux). Note: on Windows, SCHANNEL may also enforce CRL/OCSP revocation checks that can fail for internal CAs whose revocation endpoints aren't reachable from all networks — this is the `CRYPT_E_NO_REVOCATION_CHECK` error we observed with `curl`.

### Option C: Relax only for the diagnostic probe

If relaxing certificate verification globally is undesirable, make only `http_reachable` (used by the "Test Connection" diagnostic) accept non-public CAs, while keeping `http_ready` (used for health checks and window restore) strict. This keeps the spirit of Invariant #2 ("any HTTP response = reachable") while preserving stricter verification for operational probes.

## Environment

- **OS:** Windows 10
- **hermes-desktop-rust version:** 0.6.6
- **Server:** hermes-webui behind Caddy 2 with internal CA

## Verification

Tested with `ureq` 2.12.1 against the affected server:

| Configuration | Result |
|---------------|--------|
| `ureq` default (`rustls` + `webpki-roots`) | ❌ `UnknownIssuer` |
| `ureq` + `native-certs` | ❌ `UnknownIssuer` (rustls-native-certs did not load the internal CA on Windows in our testing) |
| `ureq` + `native-tls` | ❌ `UnknownIssuer` (appears to use OpenSSL fallback rather than SCHANNEL on Windows) |
| `curl` without flags | ❌ `CRYPT_E_NO_REVOCATION_CHECK` |
| `curl --ssl-no-revoke` | ✅ HTTP 302 |
| `python requests` (certifi) | ❌ `CERTIFICATE_VERIFY_FAILED` |
| Browser (Chrome/Edge) | ✅ Works |

The `native-certs` approach did not work in our testing on Windows with ureq 2.12.1 + rustls-native-certs 0.7.3 — the internal CA was present in both `Cert:\CurrentUser\Root` and `Cert:\LocalMachine\Root` but was not loaded. This may be a `rustls-native-certs` bug / limitation that warrants further investigation as part of the fix.

A working approach we verified: configuring `rustls` with a custom `ServerCertVerifier` that delegates to the platform verifier (`rustls-platform-verifier` crate), which on Windows uses the Windows certificate store properly — this successfully connected to the affected server while still performing proper certificate validation (hostname, expiry, chain).

nesquena:
## Symptom (#60, reported by ycj)

"Test Connection" reports **✗ Unreachable** for servers that load fine in any browser — whenever the server's TLS certificate chains to an enterprise/internal CA (corporate PKI, Caddy Local Authority, self-signed CA imported into the OS trust store).

## Root cause

The Rust-side probes (`health::http_reachable` / `http_ready` — used by Test Connection, the connect preflight, tunnel readiness, and health polling) build default `ureq` agents, which verify TLS against the **bundled Mozilla root store only** (`webpki-roots`). The system webview (WKWebView / WebView2 / WebKitGTK) verifies against the **OS trust store** — so the page itself loads while our probes call the same server unreachable. Enterprise CAs live exactly in that gap.

## Why not PR #59's approach

#59 disabled certificate verification entirely (`curl -k` equivalent) — that silently accepts *any* certificate for every probe, including MITM'd ones, and was rightly closed.

## Fix

Probes now verify against the **union** of the bundled Mozilla roots and the OS trust store (`rustls-native-certs`), built once and cached:

- **Strictly additive** — every certificate that verified before still verifies; a broken/empty OS store degrades to exactly the old behavior. (Deliberately NOT ureq's own `native-certs` feature, which *substitutes* the store and fails open to an empty one — all-HTTPS-fails — on load errors.)
- **Verification is never skipped or weakened.**
- Provider/protocol construction mirrors ureq's own `rtls.rs` (explicit *ring* provider, TLS 1.2+1.3) so no process-default-provider ambiguity.
- Startup log line reports webpki + OS root counts for field debugging.

## Verification

- New unit tests: union-property guard (roots can never fall below the webpki baseline) + config-build guard; plus an `#[ignore]`d live-network round-trip (passes locally against a public host through the new store).
- `cargo fmt` / `clippy --all-targets -- -D warnings` / `cargo test` (20 passed) clean; `cargo xwin check` clean (Windows uses the same pure-Rust path; `rustls-native-certs` reads SChannel/Keychain/ca-certificates per platform).
- Versions pinned to what ureq 2.x already pulls in — no new major dependency trees.

Definitive field verification needs a machine with an enterprise CA in the OS store — @ycj, if you're able to test a build of this branch against your Caddy setup, that would close the loop.

Closes #60

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,20 @@
 # Changelog
 
+## [v0.6.9]
+
+### Fixed
+
+- **"Test Connection" (and all connection probes) now trust enterprise and
+  internal CAs installed in the OS trust store.** The Rust-side probes
+  verified TLS against the bundled Mozilla root store only, so a server
+  behind a corporate PKI or an internal CA (e.g. Caddy Local Authority) that
+  the OS trusts — and that the webview itself happily loads — reported
+  "✗ Unreachable". Probes now verify against the union of the bundled roots
+  and the OS trust store: strictly additive, so every certificate that
+  verified before still does, and certificate verification is never skipped
+  or weakened. (#60, reported by ycj; supersedes the declined
+  disable-verification approach from PR #59.)
+
 ## [v0.6.8] — 2026-06-30
 
 ### Fixed
--- a/src-tauri/Cargo.lock
+++ b/src-tauri/Cargo.lock
@@ -1717,6 +1717,8 @@ dependencies = [
  "objc2-app-kit 0.2.2",
  "objc2-foundation 0.2.2",
  "regex",
+ "rustls",
+ "rustls-native-certs",
  "serde",
  "serde_json",
  "tauri",
@@ -1732,6 +1734,7 @@ dependencies = [
  "tauri-plugin-updater",
  "ureq",
  "url",
+ "webpki-roots 0.26.11",
  "x11-dl",
 ]
 
--- a/src-tauri/Cargo.toml
+++ b/src-tauri/Cargo.toml
@@ -25,6 +25,12 @@ serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 log = "0.4"
 ureq = "2"
+# Probe TLS trust (issue #60): health.rs unions the webpki baseline with the
+# OS trust store so enterprise/internal CAs verify like they do in the system
+# webview. Versions/features track what ureq 2.x already pulls in.
+rustls = { version = "0.23", default-features = false, features = ["ring", "logging", "std", "tls12"] }
+webpki-roots = "0.26"
+rustls-native-certs = "0.8"
 arboard = "3"
 image = { version = "0.25", default-features = false, features = ["png"] }
 base64 = "0.22"
--- a/src-tauri/src/health.rs
+++ b/src-tauri/src/health.rs
@@ -2,14 +2,62 @@
 //! GET (never HEAD — servers may 405/501 it), and ANY HTTP response —
 //! including 4xx/5xx — counts as reachable; only transport errors fail.
 
+use std::sync::{Arc, OnceLock};
 use std::time::Duration;
 
-pub fn http_reachable(url: &str, timeout: Duration) -> bool {
-    let agent = ureq::AgentBuilder::new()
+/// Probe TLS roots: the bundled Mozilla store UNIONed with the OS trust
+/// store. ureq's default is webpki-roots alone, which rejects
+/// enterprise/internal CAs the user's OS trusts — the system webview
+/// (WKWebView/WebView2/WebKitGTK) verifies against the OS store, so the page
+/// loads fine while Test Connection said "Unreachable" (issue #60). The union
+/// is deliberately additive-only: everything that verified before still
+/// verifies, and a broken/empty OS store can never make things worse than
+/// the webpki baseline (ureq's own `native-certs` feature REPLACES the
+/// baseline and fails open to an empty store — not acceptable here).
+/// Certificate verification itself is never skipped (the closed PR #59
+/// approach).
+fn probe_roots() -> rustls::RootCertStore {
+    let mut roots = rustls::RootCertStore {
+        roots: webpki_roots::TLS_SERVER_ROOTS.to_vec(),
+    };
+    let baseline = roots.roots.len();
+    let native = rustls_native_certs::load_native_certs();
+    for e in &native.errors {
+        log::warn!("probe tls: OS trust store: {e}");
+    }
+    let (added, unparsable) = roots.add_parsable_certificates(native.certs);
+    log::info!("probe tls: {baseline} webpki roots + {added} OS roots ({unparsable} unparsable)");
+    roots
+}
+
+fn probe_tls_config() -> Arc<rustls::ClientConfig> {
+    static CONFIG: OnceLock<Arc<rustls::ClientConfig>> = OnceLock::new();
+    CONFIG
+        .get_or_init(|| {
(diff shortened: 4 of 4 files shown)


nesquena:
# Review record (deep-review process)

**Verdict: merge.** Traced end-to-end before opening; verification receipts:

- **Approach vetted against the alternatives**: PR #59's disable-verification was rejected (accepts any cert, MITM-blind); ureq's own `native-certs` feature was rejected after reading its vendored source — it *substitutes* the OS store for the webpki baseline and degrades to an **empty root store** (all HTTPS fails) if the OS load errors. The union store is strictly additive with a tested floor.
- **Blast radius**: `health.rs` is the only probe surface (Test Connection, connect preflight, tunnel readiness, health/recovery polling). Tunnel readiness probes hit `http://localhost` — TLS path unaffected. The updater uses its own HTTP stack (tauri-plugin-updater) — untouched.
- **Version alignment**: rustls 0.23 / webpki-roots 0.26 / rustls-native-certs 0.8 match what ureq 2.12.1 already resolves in Cargo.lock; provider construction copied from ureq's `rtls.rs` (explicit *ring*, TLS 1.2+1.3) so no process-default CryptoProvider ambiguity.
- **Tests**: union-property guard (roots ≥ webpki baseline — pins the never-shrink invariant), config-build guard, `#[ignore]`d live round-trip (passed locally through the new store). 20/20 green, clippy `-D warnings` clean, `cargo xwin check` clean, Linux smoke green on this branch ([run 29171489838](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/29171489838)).
- **Outstanding**: definitive enterprise-CA field test needs a machine with an internal CA in the OS store — requested from @ycj in the PR body; the fix ships as strictly-additive so the downside risk is nil.

Issue #81, opened 2026-07-17
Approval prompt repeats until the app crashes (Linux/Wayland) — likely per-webview notification pile-up, not the WebUI card


nesquena-hermes:
## Summary

Reported by **digitwo** in Discord `#webui` (Zorin OS / Linux, Wayland, desktop app).

> "Last night I have a case when approval continue to appear until the app crash."

An approval prompt **kept re-appearing repeatedly** (a loop of approval notifications/cards) until the desktop **app crashed**. Reported without a specific repro, so filed as needs-triage with an investigation trail.

## Why this looks like a desktop-shell issue, not the WebUI approval card

I traced the WebUI's approval surface and it is structurally **incapable of "appearing repeatedly until crash"** on its own:

- The approval prompt is a **single reused DOM element** (`#approvalCard`) — `showApprovalCard()` (`static/messages.js:6947`) writes into that one card, it does not append new nodes.
- It is **signature-deduped**: `_approvalSignature` (built from desc+cmd+session+approval_id) short-circuits re-render of the same approval (`sameApproval` guard, ~6956), and dismissed approvals are suppressed via `_isApprovalDismissed` (~6950).
- Polling is a single `setInterval(_tick, 1500)` with one live timer (`_approvalPollTimer`, ~7174), guarded to the active session (`_approvalPollingSessionMissingOrMismatched`).

So a runaway, memory-growing "approval keeps appearing → crash" is far more consistent with the **desktop shell re-surfacing the approval notification** than with the web card itself. This matches the shape of existing desktop notification issues:
- **#51** — "notification re-appears in every tab on app re-open" (per-webview re-display, no cross-tab dismiss).
- Queue item #6 (Lemz) — same notification/popup re-appears in every tab; wants a universal dismiss.
- **#72** — actionable approve/deny from the OS notification (the desktop notification surface for approvals).

Each strip tab is its own WebKitGTK webview; an outstanding approval that is bridged to a native notification (or re-rendered per webview) can pile up across tabs/relaunch, and unbounded native-notification churn on WebKitGTK/Tauri can drive memory/handle exhaustion → crash.

## What I could NOT determine (needs reporter detail)

- Whether it reproduced across **multiple tabs** or a single tab (multi-tab → strongly desktop-shell / #51 family).
- Whether the repeating element was a **native OS notification**, the **in-webview approval card**, or both.
- What "crash" means: app window closed, became unresponsive, or OOM-killed.
[...]

Issue #78, opened 2026-07-17
Blank window on Fedora 44


miguelangelnieto:
I am trying to run Hermes Desktop on Fedora 44, and I get a blank window. The console displays this message:

Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

I am using the appimage, latest release. I see in the window title that it tries to connect to localhost by default, and my WebUI is remote, but since I just get a blank page I cannot change the configuration.

nesquena:
@miguelangelnieto thanks for the report — this is WebKitGTK's DMABUF renderer failing on your EGL stack, and there's an immediate workaround you can use today:

```bash
WEBKIT_DISABLE_DMABUF_RENDERER=1 ./Hermes.WebUI.Desktop_*_lin_x86_64.AppImage
```

**Shipping in v0.7.0** (today): a built-in escape hatch — `HERMES_DESKTOP_SAFE_RENDER=1` — which flips WebKitGTK to its safe rendering paths (DMABUF renderer + compositing mode off) before any webview initializes.

For your second problem (target defaults to localhost and you can't reach Preferences while the window is blank): your config lives at `~/.config/ai.get-hermes.HermesWebUIDesktop/prefs.json` — edit `targetUrl` there while the app is closed and it will connect to your remote WebUI on next launch.

Could you confirm whether the env var brings the window back on Fedora 44? If it does, we'll look at auto-detecting the EGL failure and applying the fallback automatically so nobody needs the flag.

miguelangelnieto:
@nesquena thanks a lot for your answer.

I have downloaded the lastest 0.7.0 and used the env workaround. I am still getting a blank window:

```
› WEBKIT_DISABLE_DMABUF_RENDERER=1 ./Hermes.WebUI.Desktop_0.7.0_lin_x86_64.AppImage 
[2026-07-21][12:53:49][hermes_webui_desktop::conn][INFO] conn: connecting (mode=direct, target=http://localhost:8787)
[2026-07-21][12:53:49][hermes_webui_desktop::health][INFO] probe tls: 119 webpki roots + 146 OS roots (0 unparsable)
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
[2026-07-21][12:53:50][hermes_webui_desktop::conn][WARN] conn: failed — showing error window
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
[...]

PR #82, opened 2026-07-19
v0.7.0 issue sweep: Wayland gap, mac tab-state indicators, notification dedupe, stale spinner, restore probe, refresh-all + QoL

### LINKED ISSUE #64: macOS: surface per-tab needs-attention (approval/clarify) the Mac-native way — attention ● is computed then discarded (follow-up to #14/#44)
Follow-up to **#14** (per-tab needs-attention indicator) and **#44** (the Mac-native profile-indicator precedent). Reported by **b3nw**, who runs the app on both macOS and Windows.

### What's shipped vs. what's missing

#14 shipped a per-tab "needs attention" signal (a pending approval/clarify prompt) **only on the Windows/Linux tab strip** — an amber attention dot + tint. macOS native tabs get nothing.

The interesting part: the attention `●` marker actually *arrives* at the macOS code path — it's just deliberately discarded.

In `src-tauri/src/windows.rs`, `apply_reported_title()` (L818-834):

```rust
pub fn apply_reported_title(app: &AppHandle, label: &str, raw: &str) {
    let (attention, title) = split_attention(raw);   // attention is computed
    if clean_title(title).is_empty() {
        return;
    }
    if label.starts_with("tab-") {
        strip::set_tab_title(app, label, title, attention);   // Win/Linux: uses it
    } else {
        // macOS / regular content windows: no per-tab badge, but store the
        // marker-free title so the native window title never shows a stray "●".
        app.state::<AppState>()
            .raw_titles.lock().unwrap()
            .insert(label.to_string(), title.to_string());     // attention dropped
        refresh_title(app, label);
    }
}
```

So on macOS the `attention` bool is parsed and then thrown away — the marker is stripped so the native window title stays clean, and there's no Mac-native surface for it.

### Fix shape

Same pattern as #44 (which surfaced the profile indicator Mac-native via the tab **title** prefix when the Win/Linux strip dot wasn't available). The attention signal is *already computed* here — instead of discarding it on the macOS branch, re-surface it through the native-title carrier (e.g. a small glyph prefix on the tab title for windows with a pending approval/clarify), respecting macOS conventions.

This is a clean, low-risk follow-up: the input already exists, only the macOS output is missing.

### Note on platform

b3nw uses both macOS and Windows. On Win/Linux this **is** shipped (#14), so if he's only missing it on his Mac this is the expected macOS gap. If he's also missing it on Windows that would be a separate regression. b3nw isn't reachable for confirmation (not in the project Discord; no prior activity on this repo), so **open question for the implementer**: verify from source whether the Win/Linux strip path actually surfaces this — if it does, this stays a macOS-only gap; if not, widen the fix to cross-platform.

Refs: #14, #44.

### LINKED ISSUE #65: macOS: surface per-tab active/idle (session-in-progress) the Mac-native way — BUSY_REPORTER is strip-only, not injected on macOS (follow-up to #46/#44)
Follow-up to **#46** (per-tab active/inactive "session-in-progress" indicator) and **#44** (the Mac-native profile-indicator precedent). Reported by **b3nw**, who runs the app on both macOS and Windows.

### What's shipped vs. what's missing

#46 shipped (v0.6.5) a per-tab **working spinner** on the Windows/Linux tab strip, driven by the page's `S.busy` flag. macOS native tabs get no active/idle cue at all.

The reason: the busy reporter is never injected on macOS.

In `src-tauri/src/bridge.rs`, `init_script()` (L615-621):

```rust
if cfg!(not(target_os = "macos")) {
    parts.push(SHORTCUT_FORWARDER);
    // Profile dot + working glyph are strip-only (Windows/Linux); macOS
    // native tabs have no strip, so skip these reporters there (#31, #46).
    parts.push(PROFILE_REPORTER);
    parts.push(BUSY_REPORTER);
}
```

`BUSY_REPORTER` (the script that watches `S.busy` and emits `busy` over the bridge — defined L549) is gated behind `cfg!(not(target_os = "macos"))`, so it's simply not present on macOS. With no `busy` events arriving, there's nothing to drive a native cue even if one existed.

### Fix shape

Same pattern as #44 — the profile indicator was strip-only until it got a Mac-native equivalent via the tab title. Two pieces here:

1. Inject `BUSY_REPORTER` on macOS too (or a macOS variant) so `busy` events flow.
2. Surface the active/idle state Mac-native — e.g. a working glyph on the native tab title, riding the same title carrier #44 established.

### Note on platform

b3nw uses both macOS and Windows. On Win/Linux this **is** shipped (#46), so if he's only missing it on his Mac this is the expected macOS gap. If he's also missing it on Windows that would be a separate regression. b3nw isn't reachable for confirmation (not in the project Discord; no prior activity on this repo), so **open question for the implementer**: verify from source whether the Win/Linux strip path actually surfaces this — if it does, this stays a macOS-only gap; if not, widen the fix to cross-platform.

Refs: #46, #44.

### LINKED ISSUE #74: Desktop: tab "working" spinner goes stale on inactive tabs — stays spinning after session finished (clears on focus)
## Summary

On the Windows/Linux tab strip, a tab's **"working" spinner goes stale** — it keeps showing the session as in-progress after the session has actually finished (minutes ago). Clicking/focusing the tab makes it immediately update to the correct done state.

Reported by **b3nw** on **v0.6.9** (Discord).

> sessions regularly show that they are "working" until you click on them and then they were actually done minutes+ ago.

## Steps to reproduce

1. Open the desktop app with two or more tabs.
2. Start a session running in a **non-focused** tab (or switch away while it runs).
3. Let the session finish while that tab stays in the background.
4. Observe the tab strip still shows the working spinner for that tab.
5. Click the tab — it immediately flips to the done state (it had actually finished minutes earlier).

## Expected

The per-tab working indicator clears when the session finishes, regardless of whether the tab is currently focused.

## Suspected area

The per-tab working spinner (shipped #46, v0.6.5) is driven by `BUSY_REPORTER` watching the page's `S.busy` flag and emitting `busy` events over the bridge (`src-tauri/src/bridge.rs`). The "click and it's already done" recovery points at **the background tab's webview being throttled** — while backgrounded, WebView2/WKWebView suspends timers/JS so the `busy=false` transition isn't emitted until the tab is foregrounded and the webview resumes.

**Cross-repo ambiguity** (same shape as #56/#65): could also be that the WebUI page's `S.busy` itself only settles on a focus/visibility event rather than driving off the SSE completion. Lean desktop-shell throttling given the focus-fixes-it symptom, but confirm which layer stops emitting.

## Note

Distinct from **#65** (that one is macOS-specific — `BUSY_REPORTER` is never injected on macOS, so there's no cue at all). This is the Windows/Linux cue existing but going **stale** on inactive tabs.

---
_Filed from a Discord community report. Edit, re-route, or close if mis-scoped._

### LINKED ISSUE #75: feat: click-to-copy the version number in the ⋮ overflow menu (for bug reports)
## Summary

Make the version number in the **⋮ (three-dots) overflow menu** click-to-copy, so it's easy to paste when writing a bug report.

Requested by **b3nw** on **v0.6.9** (Discord).

> FR: make the version number in the 3 dots menu click -> copy for easy pasting when bug reporting 🙂

## Proposed behavior

- In the overflow menu where the version is shown, make the version row a click target that copies the version string to the clipboard.
- Ideally copy a full, paste-ready identifier — the desktop app version plus the underlying WebUI version (e.g. `desktop 0.6.9 / webui v0.52.x`) — since bug reports usually need both.
- Brief "Copied" affirmation (toast / transient checkmark) on click.

## Why

Streamlines community bug reports — reporters can grab an exact, unambiguous version string without retyping it. Low-risk, small, purely additive.

---
_Filed from a Discord community report. Edit, re-route, or close if mis-scoped._

### LINKED ISSUE #76: Desktop: after a web-app update every tab shows a "must hard refresh" message needing per-tab clicks — offer auto-refresh of inactive tabs and/or a "refresh all"
## Symptom
After the underlying web app updates, every open desktop tab shows a "must hard refresh" message that has to be dismissed/actioned individually, one tab at a time (extra clicks per tab). The reporter requests that inactive tabs be auto-refreshed and/or that a single "refresh all" option be provided.

## Platform
Desktop app (wrapper), multi-tab strip.

## Suspected area
AMBIGUOUS. The hard-refresh banner is WebUI-rendered on a client/server version mismatch (WebUI surface), while coalescing or auto-applying it across the desktop strip's per-tab webviews is a shell concern (desktop shell).

## Relationship to #56
Same cross-tab coordination family as #56, but a **distinct symptom at a different lifecycle stage**:
- **#56** = the *pre-update upgrade prompt* re-appears in every tab (no coordination when one tab initiates the upgrade).
- **This** = the *post-update "must hard refresh" banner* that then requires per-tab dismissal, with no auto-refresh of inactive tabs and no "refresh all".

A shell-level cross-tab upgrade/refresh coordinator would likely address both; filing separately so the post-update per-tab refresh burden + the "refresh all" ask aren't lost inside #56.

## Repro
1. Have multiple desktop tabs open.
2. Update the underlying web app.
3. Observe each tab now shows a "must hard refresh" message.
4. Each must be handled per-tab; there is no auto-refresh of inactive tabs and no "refresh all".

## Requested
- Auto-refresh inactive tabs (where safe), and/or
- A single "refresh all" action.

---
_Filed from a Discord community report. Edit, re-route, or close if mis-scoped._

### LINKED ISSUE #80: Wayland: huge empty space at top of window (X11-only #67 scale fix applied unconditionally to Wayland)
## Summary

Reported by **digitwo** in Discord `#webui` (Zorin OS, **Wayland** session, Linux desktop app).

> "On Linux (I use Zorin, which run with Wayland), the desktop app have display problem: there are a huge space on the top."

The WebUI content is pushed down by a large empty band at the top of the window, under the tab strip — the same visual symptom as the previously-fixed **#67** ("Linux: large empty gap above the WebUI under the tab strip"), but on a **Wayland** session where the #67 fix may not apply (or may over-correct).

## Why this is likely NOT covered by the #67 fix

#67 was root-caused to **X11 only**: wry's WebKitGTK backend re-derives its own scale factor for the child webview from the **X screen's reported physical size** (`scale_factor_from_x11 = screen_px * 25.4 / screen_mm / 96`), which is bogus on VMs/Xvfb/remote-X, so a logical 38px strip offset got multiplied into ~171 physical px. The fix (`strip.rs` `child_position` / `child_size`) compensates by passing **physical** coordinates for **all** Linux:

```rust
fn child_position(x: f64, y: f64, scale: f64) -> Position {
    if cfg!(target_os = "linux") {
        PhysicalPosition::new((x * scale).round() as i32, (y * scale).round() as i32).into()
    } else {
        LogicalPosition::new(x, y).into()
    }
}
```

The compensation is gated on `cfg!(target_os = "linux")` — it does not distinguish **X11 vs Wayland**. On Wayland there is no X screen, so WebKitGTK does not perform the bogus X11-physical-size scale derivation; the compositor hands the correct scale to the webview directly. Multiplying our logical 38px offset by `win.scale_factor()` again on Wayland can therefore **double-apply** the scale factor (e.g. on a HiDPI/fractional-scale Wayland display, `38 * 2 = 76` physical px positioned into a surface that already accounts for scale), pushing the content webview down — the "huge space on the top" the reporter sees.

So this reads as a **Wayland regression/blind spot of the #67 X11 fix**: the fix corrects X11's DPI lie but is applied unconditionally to Wayland where the premise (wry re-derives a wrong scale) doesn't hold.

## Investigation notes

- `main.rs` already special-cases Wayland-vs-X11 at startup (dlopen'ing libX11 for `XInitThreads`, skipped on "Wayland-only systems without libX11"), so the shell knows the distinction exists — but `strip.rs child_position/child_size` do not consult it.
- The headless smoke test runs against a fake X screen, so neither the original X11 bug nor a Wayland over-correction is caught in CI.

## Suggested fix shape

Make the Linux child-webview coordinate transform **backend-aware**:
- Detect the session type (e.g. `XDG_SESSION_TYPE=wayland` / `GDK_BACKEND`, or whether libX11 loaded).
- On **X11**, keep the #67 physical-coordinate compensation (needed for the DPI lie).
- On **Wayland**, use the coordinates the compositor/tao scale expects without the extra X11-motivated multiply (likely logical, or physical computed from the Wayland scale rather than re-multiplying), so the strip offset lands at the real 38px.

Needs verification on a real Wayland session at scale 1 and at fractional/HiDPI scale.

## Steps to reproduce
1. Run the Linux desktop app on a Wayland session (e.g. Zorin OS / GNOME Wayland).
2. Observe a large empty band at the top of the window, above the WebUI content, under the tab strip.

## Info to confirm from reporter
- Desktop app version (⋮ overflow menu → version).
- Display scale (100% / fractional / HiDPI) and monitor setup.
- Whether launching under X11 (`GDK_BACKEND=x11` / an Xorg session) makes the gap disappear — this would confirm the Wayland-specific path.

## Environment
- OS: Zorin OS (Linux)
- Session: **Wayland**
- Prior related (X11, CLOSED): #67

nesquena:
Nine tracker items in one sweep. Per-issue details in the CHANGELOG entry and commit message; the high-risk items and their verification:

## Fixes

- **#80 Wayland top gap** — the #67/#68 physical-coordinate compensation exists to defeat wry's **X11-only** bogus DPI derivation, but applied on all Linux; on native Wayland it over-corrects into the reported gap. Now gated on GDK actually running X11 (`x11_from_env`, ordered-preference-list semantics incl. XWayland, unit-tested). X11 behavior byte-identical — Linux smoke (X11/Xvfb) re-validates the #67 fix still holds.
- **#51/#81 notification dedupe** — all tabs funnel through the single `"notify"` bridge arm; identical title+body now suppressed within a fixed 30s window (insert-on-show only → a persistent repeater reminds once per window instead of flooding until crash). Pure `dedupe_decide` unit-tested. #81's page-side re-fire root cause on Wayland remains open; this caps the blast radius natively.
- **#74 stale spinner** — background webviews throttle timers, silencing BUSY_REPORTER; the existing 4s native loop now evals `__hermesReportBusy()` into every content webview (evals ignore timer throttling; debounce keeps it emit-on-change).
- **#37 restore probe (round 3)** — the v0.6.1 retry only caught the root-bounce variant; a one-shot in-page probe ~5s post-create now reloads a restored tab showing the webui's exact in-place failure marker (`sessions.js:1862` string).

## Features

- **#64/#65 macOS per-tab state** — native tab titles adorn with ● (attention — previously computed then dropped in `apply_reported_title`) / ⟳ (busy — BUSY_REPORTER now injected on macOS), same mechanism as the #44 profile prefix. Attention outranks busy.
- **#76 Refresh All Tabs** (⋮ menu, mac View menu, Ctrl/Cmd+Shift+R) — one action clears per-tab hard-refresh banners after a WebUI update.
- **#75 click-to-copy version**; **#78 `HERMES_DESKTOP_SAFE_RENDER=1`** (WebKitGTK DMABUF/compositing off before webview init — EGL blank-window recovery; explicit `WEBKIT_*` always wins).

## Verification

- 23 unit tests green (new: GDK backend choice, dedupe window semantics); clippy `-D warnings` clean; `cargo xwin check` clean.
- **Live macOS run** (`HERMES_TEST_TAB_AFTER`): 2 native tab attaches, 8/8 main-thread heartbeats, 3 windows, zero panics — exercises the changed init-script assembly on every mac webview.
- Linux smoke dispatched on this branch (X11 path regression check for #80's gate).
- Field confirmation requested from reporters: digitwo (#80 Wayland, #81 flood), b3nw (#74, #37), miguelangelnieto (#78 flag).

Closes #64
Closes #65
Closes #74
Closes #75
Closes #76
Closes #80

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,65 @@
 # Changelog
 
-## [v0.6.9] — 2026-07-11
+## [v0.7.0] — 2026-07-18
+
+An issue-sweep release: nine tracker items land at once — Wayland layout,
+macOS per-tab state indicators, cross-tab notification dedupe, stale-spinner
+and restored-session fixes, refresh-all, and quality-of-life items.
+
+### Added
+
+- **macOS: native tabs now show per-tab state** — "●" prefixed to the tab
+  title when a session is waiting on you (approval/clarify), "⟳" while it's
+  working (issues #64/#65). The signals existed (the page reports both) but
+  macOS discarded them; they now adorn the native tab title the same way the
+  profile-name prefix does, so a backgrounded tab's state is visible at a
+  glance. Attention outranks working when both apply.
+- **Refresh All Tabs** (issue #76): one action — in the ⋮ menu, the macOS
+  View menu, or Ctrl/Cmd+Shift+R — reloads every tab in every window, so a
+  WebUI update no longer means clicking the "must hard refresh" banner in
+  each tab individually.
+- **Click-to-copy version** (issue #75): the version row in the ⋮ menu (and
+  a new "Copy Version" item in the macOS app menu) copies a paste-ready
+  identifier for bug reports, confirmed with a notification.
+- **Linux safe-render escape hatch** (issue #78): launching with
+  `HERMES_DESKTOP_SAFE_RENDER=1` disables WebKitGTK's DMABUF renderer and
+  compositing mode before any webview initializes — the recovery path for
+  GPU/driver combos where EGL fails and the window comes up blank (e.g.
+  "Could not create default EGL display: EGL_BAD_PARAMETER" on Fedora).
+  Explicitly-set `WEBKIT_*` variables always win.
+
+### Fixed
+
+- **Wayland: huge empty space above the WebUI** (issue #80). The v0.6.8 fix
+  for the X11 gap (#67) passes physical child-webview coordinates to defeat
+  wry's bogus X11 DPI derivation — but it applied on ALL Linux, and on native
+  Wayland (where wry's scale is sane) it over-corrects into a top gap. The
+  compensation now applies only when GDK actually runs its X11 backend
+  (including XWayland, which needs it); native Wayland gets plain logical
+  coordinates.
+- **Duplicate notifications across tabs, and notification floods** (issues
+  #51/#81). Every tab observes the same server events, so N open tabs meant
+  N identical OS notifications — and a page re-firing for a still-pending
+  approval could pile up notifications indefinitely (reported on Linux/
+  Wayland as "approval kept appearing until the app crashed"). All tabs
+  funnel through one native handler, which now suppresses identical
+  notifications within a 30-second window: many tabs produce one
+  notification, and a persistent repeater surfaces at most once per window
+  as a gentle reminder instead of a flood.
+- **Tab "working" spinner no longer goes stale on background tabs** (issue
+  #74). Hidden webviews throttle their timers, so the in-page busy reporter
+  could go quiet and leave a finished session spinning until you clicked the
+  tab. The shell now re-polls every tab's busy state on its own 4-second
+  loop — an eval runs regardless of timer throttling — so the spinner clears
+  within seconds of the session finishing, focused or not.
+- **Restored session tabs showing "Session not available in web UI." at
+  launch** (issue #37, third and hopefully final round). The v0.6.1 retry
+  only caught the variant where the WebUI bounces the tab back to the root
+  URL; the reported symptom shows the message in place at the deep URL. The
+  shell now also probes the restored tab's page once (~5s after creation)
+  and reloads it in place if the marker is present — the programmatic
+  equivalent of the switch-away-and-back that always fixed it. Healthy tabs
+  match nothing and pay nothing.
 
 ### Fixed
 
--- a/src-tauri/src/bridge.rs
+++ b/src-tauri/src/bridge.rs
@@ -412,6 +412,7 @@ const SHORTCUT_FORWARDER: &str = r##"
     var send = function (v) { e.preventDefault(); e.stopPropagation(); EMIT('shortcut', v); };
     if (k === 'tab') { send(e.shiftKey ? 'prev-tab' : 'next-tab'); return; }
     if (k === 'b' && e.shiftKey) { send('toggle-bar'); return; }
+    if (k === 'r' && e.shiftKey) { send('reload-all'); return; }
     if (e.shiftKey) return;
     if (k === 't') send('new-tab');
     else if (k === 'n') send('new-window');
@@ -559,6 +560,12 @@ const BUSY_REPORTER: &str = r##"
     else report();
     setInterval(report, 700);
     window.addEventListener('focus', report);
+    // Native re-poll hook (issue #74): hidden webviews throttle/suspend timers,
+    // so a session finishing in a background tab may never tick the interval —
+    // the spinner goes stale until focus. The shell evals this on its own 4s
+    // loop; an eval executes even when timers are throttled, and it re-reads
+    // the live S.busy at that moment.
+    window.__hermesReportBusy = report;
   })();
 "##;
 
@@ -614,11 +621,13 @@ pub fn init_script(label: &str, pre_paint_hex: &str, is_ssh: bool) -> String {
     }
     if cfg!(not(target_os = "macos")) {
         parts.push(SHORTCUT_FORWARDER);
-        // Profile dot + working glyph are strip-only (Windows/Linux); macOS
-        // native tabs have no strip, so skip these reporters there (#31, #46).
+        // Profile dot is strip-only (Windows/Linux); macOS gets its profile
+        // indicator from MACOS_PROFILE_REPORTER above (#31, #44).
         parts.push(PROFILE_REPORTER);
-        parts.push(BUSY_REPORTER);
     }
+    // Busy state feeds the strip spinner on Win/Linux (#46) AND the native
+    // tab-title "⟳" adornment on macOS (#65) — inject everywhere.
+    parts.push(BUSY_REPORTER);
     parts.push(THEME_BRIDGE);
     parts.push(ROUTE_REPORTER);
     parts.push(WINDOW_OPEN);
@@ -688,11 +697,15 @@ pub fn install(app: &AppHandle) {
                     crate::strip::set_tab_dot_profile(&handle, &label, &name);
                 }
             }
-            // Tab busy/streaming state for the per-tab "working" glyph (#46).
+            // Tab busy/streaming state: the per-tab "working" glyph on the
+            // Win/Linux strip (#46), and the "⟳" native-title adornment on
+            // macOS content windows (#65).
             "busy" => {
+                let busy = payload["value"].as_bool().unwrap_or(false);
                 if crate::strip::enabled() && label.starts_with("tab-") {
-                    let busy = payload["value"].as_bool().unwrap_or(false);
                     crate::strip::set_tab_busy(&handle, &label, busy);
+                } else if cfg!(target_os = "macos") && label.starts_with("main-") {
+                    windows::set_window_indicator(&handle, &label, Some(busy), None);
                 }
             }
             // Active-profile NAME for the macOS native tab title (issue #44).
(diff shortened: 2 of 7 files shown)


nesquena:
# Review record (deep-review process)

**Verdict: merge.** Receipts beyond the PR body:

- **#80 regression safety**: the smoke harness runs under Xvfb = X11, so its green run ([29670910254](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/29670910254)) proves the new gate keeps the X11 physical-coordinate path byte-identical — the #67 gap fix still renders correctly. The Wayland leg can't be CI-validated; digitwo is the field check.
- **#51/#81 dedupe placement**: verified all tabs funnel through the single `listen_any("bridge")` notify arm (no second notification path — the DOM watcher was removed in v0.6.6), so one cache covers every source. Fixed-window semantics unit-tested including the it-must-not-slide property.
- **#74 mechanism check**: eval-vs-timer-throttling is the load-bearing assumption; the reporter's debounce means the 4s poke emits only on change, and the poke rides the existing menu-guarded loop (no new #33-class modal re-entry — same guard skips the tick while the ⋮ popup is up).
- **#64/#65**: adornment applied in `refresh_title` OUTSIDE truncation (same rule as the #44 prefix); `forget()` cleans the new map; live mac run exercised the changed init-script on every webview (2 attaches, 8/8 heartbeats, 0 panics).
- **#37 probe**: marker string verified against webui `sessions.js:1862` verbatim; one-shot; runs only for deep-URL restored tabs.
- Gates: 23 tests, clippy -D warnings, xwin, all clean.

PR #84, opened 2026-07-26
fix(linux): correct X11 child webview packing

### LINKED ISSUE #83: Huge Gap at top of Window (half the window size) still not fixed (Linux x11)
## Summary
I'm experiencing similar behavior to that described in **#67** ("Linux: large empty gap above the WebUI under the tab strip") and **#80** ("Wayland: huge empty space at top of window (X11-only #67 scale fix applied unconditionally to Wayland)"). It doesn't seem to be fixed as of Version 0.7.0, which I am running. I'm on Ubuntu x11 (not Wayland).

Image

## Environment
- OS: Ubuntu (Linux)
- Session: X11
- Prior related (both CLOSED): #67, #80

albnavper:
## Summary

- fix the Linux/X11 half-window gap by correcting GTK packing for Tauri child webviews
- keep the 38px tab strip fixed and let only the content webview expand
- scope the GTK override to the X11 backend; Wayland and non-Linux platforms are unchanged
- add a regression test and changelog entry

## Root cause

On Linux, Tauri 2.11 creates `WindowChild` webviews with `build_gtk(default_vbox)`. Wry packs each child into the GTK box with `expand=true`, so the tab strip and content webview split the available height even though the app requests explicit bounds. The apparent empty band is the strip webview receiving roughly half of the window.

## Verification

- reproduced on Linux Mint Cinnamon/X11 with v0.7.0
- before: content occupied only the lower half of a 1920x1052 window
- after: the strip remains 38px high and content starts immediately below it
- `cargo fmt --check`
- `cargo test` — 24 passed, 1 ignored, 0 failed
- Linux `cargo clippy --all-targets` passes after excluding existing platform-only dead-code and uninlined-format warnings; the patch introduces no new clippy findings

Closes #83


nesquena-hermes:
## Review — Codex gate SHIP AS-IS; one caveat (couldn't compile locally)

Thanks @albnavper — this is a clean, well-scoped fix for the recurring X11 half-window gap (#83, after #67/#80). I ran it through the full advisor gate and traced the change:

**What's right:**
- **Root cause is correctly identified.** Tauri 2.11 packs Linux child webviews into a GtkBox with `expand=true` and ignores their requested bounds, so the 38px strip and content webview split the height 50/50. Correcting the GTK child packing directly (fixed strip, expanding content) is the right layer.
- **Cleanly gated.** `apply_linux_child_packing` is `#[cfg(target_os = "linux")]` and guarded by `linux_backend_is_x11()`, so macOS/Windows and Wayland are untouched — important given #80 was a regression from applying an X11 fix unconditionally to Wayland. Errors are logged, never panic.
- **The pure decision is unit-tested** (`linux_child_packing` → `(false, 38)` / `(true, -1)`), and the GTK-side glue is thin.
- **No new dependency surface:** `gtk = "0.18"` is added to `Cargo.toml`, but `gtk 0.18.2` was *already* a transitive dependency (via tauri/wry) in `Cargo.lock` on `main` — this only promotes an existing indirect dep to direct. Zero new supply-chain.

**Codex advisor verdict: SHIP AS-IS.** It verified the exact gtk-rs 0.18.2 / Tauri 2.11.2 / wry 0.55.1 signatures (`with_webview` closure, `native.inner()`, `downcast::<gtk::Box>`, `set_child_packing` arg order, `PackType::Start`, `queue_resize`), main-thread dispatch, both creation paths (strip + each content tab), and that the fixed-strip/expanding-content behavior follows documented GtkBox packing + size-request semantics and persists through GTK resize/show/hide. Only a MARGINAL note: PR CI is macOS-only, so Linux native linking + live X11 allocation aren't exercised on PRs.

**One caveat before merge:** I can't compile-verify this in my environment (no `gtk+-3.0` / `webkit2gtk-4.1` system libs here), so the "it builds + the gap is actually gone" confirmation should come from a real Linux/X11 build. Given the MARGINAL CI gap Codex flagged, the cleanest path is to run the Linux smoke (or a local `cargo build` + a quick X11 launch to eyeball the strip height) before merging — that closes the one thing neither the advisor nor I could exercise. The code review itself is clean; it's purely the native build/visual confirmation that's outstanding.

Happy to merge once a Linux build confirms it compiles and the gap is resolved.

*(Reviewed at head `0b58fc11`. Codex advisor gate + my own dependency/API/scoping trace. No merge action taken.)*

nesquena-hermes:
## Correction + addendum to my review above — the X11 gate is likely the one thing to change

My first comment ran before the full advisor analysis finished, and I need to correct a point I got backwards. I praised the `linux_backend_is_x11()` scoping as a clean virtue ("Wayland correctly untouched"). The completed gate (**verdict: SHIP WITH THE HIGH-VALUE FIXES**) found that scoping is actually the weak spot, and the reasoning is source-verified and convincing:

**The packing bug is backend-independent, so gating the fix to X11 leaves native Wayland broken with the identical 50/50 split.** (`src-tauri/src/strip.rs:41`)

- The `GtkBox pack_start(expand=true)` path that causes the strip/content to split the height has **no backend conditionality** in tauri-runtime-wry/wry — a native-Wayland session must exhibit the same bug today.
- This is **not** a reverse-#80 risk. #80 was an *X11-specific DPI coordinate compensation* over-correcting on Wayland — it was about coordinate/scale math. This fix contains **no coordinates or scale math at all**; it only corrects *which child expands*, which is equally correct on both backends.
- Supporting evidence the advisor verified from source: `set_bounds`/`set_size`/`set_position` are **complete no-ops** for GtkBox-packed children in wry 0.55.1 (`is_in_fixed_parent=false`, `x11=None`). So the earlier coordinate-based fixes (#67, #80) never operated on the real mechanism — which is precisely why this bug keeps recurring (#67 → #80 → #83). Gating this one on X11 sets up a future "#85: Wayland — WebUI fills lower half."

### Recommended change
Delete the `if !linux_backend_is_x11() { return; }` early-return from **`apply_linux_child_packing`** (keep the X11 guard in `child_position` / `child_size`, where coordinate math genuinely differs by backend). That fixes X11 *and* Wayland with the same correct, coordinate-free packing change.

If you'd rather validate on a live Wayland session before widening scope, the acceptable alternative is: ship as-is (X11-only) **but open the Wayland follow-up issue now** and note the deliberate scoping in the changelog entry — so it's a tracked, intentional decision rather than a silent gap.

Caveat: the Wayland breakage is **source-verified mechanism analysis, not runtime-verified** (no Wayland session available to the advisor or me). So the cleanest path is probably to drop the gate and have a Wayland user confirm, since the mechanism argument is strong and the change is coordinate-free.

Everything else in my first comment stands — the API usage, thread-safety, scoping of the *coordinate* helpers, no-new-dependency, and unit test are all correct (the advisor marked the content-tab packing call and the frozen-behavior unit test as harmless SKIPs). The one substantive item is the X11 gate on `apply_linux_child_packing`. Sorry for the whiplash on that point — the fuller mechanism analysis flipped my read.

*(Full advisor gate at head `0b58fc11`. Still no merge action taken; a Linux build to confirm compile + strip height remains the other open item.)*

nesquena-hermes:
Closing as superseded by #87, now merged in `0724f63`.

@albnavper, your commit went in **unchanged and under your authorship**: #87 carried `7084221` ("fix(linux): correct X11 child webview packing") verbatim, and the merge was a merge commit rather than a squash, so it's attributed to you in `main`'s history. #87 only added the follow-up from review, applying the packing on Wayland too instead of gating it to X11. Thank you for finding and fixing this one.

PR #87, opened 2026-08-27
fix(linux): correct child webview packing on X11 and Wayland

### LINKED ISSUE #83: Huge Gap at top of Window (half the window size) still not fixed (Linux x11)

1vecera:
Fixes #83 by correcting Tauri 2.11's Linux `GtkBox` packing for both child WebViews. The tab strip now has a fixed 38 px allocation, while only the active content WebView expands into the remaining height. Backend-specific coordinate compensation stays X11-only; the packing fix applies to both X11 and Wayland.

This carries the first commit from #84 unchanged, preserving @albnavper's authorship. The follow-up removes the X11-only gate requested in maintainer review and updates the regression test and changelog to cover both Linux backends.

## Verification

- `cargo fmt --check`
- `cargo test` — 24 passed; 1 explicitly ignored live-network sanity test
- `cargo clippy --all-targets -- -D warnings -A dead-code` — passed; the unmodified warning policy still reports six pre-existing Linux-only dead-code warnings outside this diff
- `cargo build --release`
- Offscreen X11 pixel smoke at 1400×900 — connected and stable; content begins immediately below the 38 px tab strip
- Native GNOME Wayland smoke on Fedora 44 with GTK 3.24.52 and WebKitGTK 2.52.5 — connected and stable; live accessibility geometry reports the strip at `y=47, h=38` and the content WebView immediately after it at `y=85, h=1083`

## Review focus

Please verify that the direct `GtkBox` child-packing override is an acceptable bounded workaround for the current Tauri/wry container behavior, and that keeping only coordinate conversion behind the X11 gate matches the intended backend split.


nesquena-hermes:
## Exact-head review at `8f2e814a2` — source clean; Linux PR gate required before merge

This is the corrected canonical successor to #84.

### Source findings

I verified the workaround against the exact locked upstream implementation:

- `tauri-runtime-wry 2.11.2` sends Linux `WindowChild` webviews through `build_gtk(window.default_vbox())`.
- `wry 0.55.1` then calls `GtkBox.pack_start(webview, true, true, 0)`, so every child expands and requested child bounds do not control Box allocation.
- `gtk = "0.18"` promotes the same gtk-rs family already present transitively in the lockfile; the PR adds no second GTK major/family.
- `native.inner()` is the actual WebKitGTK widget, its direct parent is the default Box in this path, and `set_child_packing(child, expand, true, 0, Start)` is the right API/order.
- strip: `vexpand=false`, fixed height 38, Box expand=false;
- content: `vexpand=true`, height request reset, Box expand=true;
- hidden content webviews do not consume visible Box allocation; GTK reflows on show/hide/window resize.
- coordinate compensation remains gated to X11 in `child_position` / `child_size`; packing is coordinate-free and correctly applies to native Wayland too.
- failures log rather than panic.

The pure tuple test is correct but does not exercise GTK. The first commit preserves @albnavper as author with @1vecera as committer; the second commit removes #84’s incorrect X11-only gate. #87 should be the canonical vehicle, and #84 can close as superseded only after this head is actually validated/landed, retaining credit to both contributors.

### Remaining gate

There are **no GitHub checks on this head**. The normal PR workflow runs only on `macos-14`; all new production code and the direct GTK dependency are `cfg(target_os = "linux")`, so that workflow cannot compile or link them. `.github/workflows/linux-smoke.yml` runs only on pushes to `main` or manual base-repo dispatch, which means its first automatic run would be after merge.

The PR body’s X11/Fedora Wayland results are valuable self-reported evidence, but no attached log/screenshot or exact-head hosted receipt is available to independently verify them.

Before merge, add/run a Linux pull-request gate on this exact head:

1. Ubuntu 22.04 with the existing WebKitGTK/GTK packages: `cargo fmt --check`, `cargo test`, `cargo clippy --all-targets -- -D warnings -A dead-code`, and `cargo build --release`.
2. Run the existing Xvfb smoke before merge and retain screenshot + `app.log`; add a discriminating geometry assertion that the strip is ~38 px and content begins immediately below it instead of only checking that the process stays alive.
3. Attach the claimed native Wayland geometry/log (and a screenshot) at this SHA; cover 100% plus fractional/HiDPI if available. Xvfb cannot prove Wayland.
4. Keep issue #83 open until a containing build is reporter-retested on the original Ubuntu X11 path. #80’s old coordinate regression and this Box-packing bug are different layers.

Static threat scan is CLEAN. I did not bare-execute contributor Cargo/build scripts without an external-repo sandbox.

**Verdict: source clean, changes/evidence needed before merge.** The required change is CI/proof, not a rewrite of the GTK patch. No branch mutation or merge from this review lane.

nesquena-hermes:
## Linux gate now run on this exact head — CI green, and the geometry claim is verified

My last review said the source was clean and the only thing outstanding was that **no Linux gate had ever run on this head** (the PR workflow is `macos-14` only, and `linux-smoke.yml` triggers on pushes to `main`, so its first automatic run would have been *after* merge). I've now closed that gap myself rather than leaving it on you.

I pushed `8f2e814a2` to a temporary base-repo branch, ran `linux-smoke.yml` against it via `workflow_dispatch`, and deleted the branch afterwards. The run is public and pinned to this PR's exact head:

**[Run 35682239787](https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/35682239787)** — `completed / success`, `headSha 8f2e814a2bf1306a01b043f5ce99f5924b0b2c67`

```
JOB smoke [success]
   - Install system dependencies: success
   - Install Rust: success
   - Build (debug): success          <- the gtk 0.18 dep + cfg(linux) code compile and link
   - Start fake hermes-webui on :8787: success
   - Launch app under Xvfb, assert connected, screenshot: success
   - Upload screenshot + log: success
```

So the `gtk = "0.18"` dependency resolves against the existing lockfile family, the new `cfg(target_os = "linux")` code compiles, and the app launches and connects on Ubuntu 22.04 / WebKitGTK.

### The discriminating geometry check I asked for

I also said the smoke only proves the process stays alive, and asked for an assertion that the strip is ~38px with content immediately below rather than the 50/50 split. I pulled the run's screenshot artifact and measured it directly:

| | |
|---|---|
| Screenshot | 1400×900 |
| Window content starts | y=35 |
| Strip band | y=35 → 58 (**~38px** allowing for the Xvfb window border) |
| Page content | begins at y=58 and extends to the bottom edge |
| 50/50 split | **absent** |

An independent visual read of the same PNG confirms it: a slim single-row tab strip (`Smoke chat` tab, `+`, a green connected dot, `localhost:8787`), a separator, then the fake-WebUI page filling the entire remaining height. No half-and-half division, no blank lower region. `app.log` corroborates the healthy path:

```
strip: window main-1 built, strip webview added
strip: window main-1 positioned
strip: tab tab-1-1 added to main-1
strip: window main-1 ready
conn: connected
```

That's the packing bug demonstrably fixed on a real Linux session, not just argued from source.

### What's still open — and it is not yours to close

Two caveats, stated plainly so the record is accurate:

1. **This is X11 under Xvfb, not native Wayland.** Xvfb cannot prove the Wayland path. The mechanism argument for why the fix is backend-independent (the `pack_start(expand=true)` path has no backend conditionality in tauri-runtime-wry / wry 0.55.1) is strong and I still believe it, and this PR correctly removed #84's X11-only gate on `apply_linux_child_packing`. But "verified on Wayland" would need a Wayland session, and I don't have one. Your reported Fedora Wayland result stands as the evidence there.
2. **Issue #83 should stay open** until a containing build is re-tested by the original reporter on their Ubuntu X11 setup, per my earlier note.

Neither is a change request against this branch.

### Status

Source clean at `8f2e814a2`, Linux build + Xvfb smoke green on this exact head, strip geometry verified from the run's own artifact. From my side the review is complete and this is **ready to merge** — I have not merged it, since the Wayland confirmation and the #84-supersede/credit decision are maintainer calls. Credit stays with @albnavper (original author of the packing fix) and @1vecera (removed the X11 gate, making this the canonical vehicle).

Worth folding the geometry assertion into `linux-smoke.yml` itself so future PRs get it automatically instead of relying on someone measuring the artifact by hand — happy to open that separately.

_Gate: reviewed at exact head `8f2e814a2bf1306a01b043f5ce99f5924b0b2c67`. Evidence is the hosted CI run above plus my own pixel measurement and visual read of its uploaded screenshot and `app.log`. Temporary CI branch removed; the run and artifact persist._

nesquena-hermes:
Merged in `0724f63` (merge commit, so both authors keep their commits: @albnavper's original packing fix `7084221` and @1vecera's Wayland follow-up `8f2e814`). Thanks to both of you.

Final gate at exact head `8f2e814a2bf1306a01b043f5ce99f5924b0b2c67` before merging:

- **Codex (gpt-5.6-sol): SHIP AS-IS.** One marginal, non-blocking note. The comment at `src-tauri/src/strip.rs:1133` says Linux window resizes don't refit webviews. With the new `GtkBox` packing they do reflow automatically; what's skipped is only the explicit geometry calls. I'll correct that wording in a follow-up, since it's a comment and not behaviour.
- **Linux build + Xvfb smoke** on this exact head (run 35682239787, from my earlier comment): green, with the ~38px strip / full-height content geometry measured from the run's own screenshot.
- `test` check green, `CLEAN`/`MERGEABLE`, base current with `main`.

Two things I'm keeping honest on the record:

1. Wayland is still verified only by @1vecera's Fedora report plus the mechanism argument. Xvfb exercises X11 only.
2. The PR body's `Fixes #83` auto-closed #83. I'm reopening it, as I said I would, until the original reporter confirms on a release that contains this merge.

PR #88, opened 2026-08-29
fix: persist macOS WebUI authentication

cloow:
## Summary

- keep the first restored macOS content view on WebKit's persistent data store so the WebUI `hermes_session` cookie survives app restarts
- keep later restored views incognito for per-tab profile isolation, but seed only exact-origin, destination-scoped cookies before navigation
- explicitly clear a stale `hermes_profile` cookie when restoring the default profile
- carry cookie donors across all restored native window groups without writing auth tokens into the JSON session blob
- preserve direct plain-HTTP WebUI targets on localhost, Tailscale, LAN, and SSH forwards via the scoped macOS web-content ATS exception

## Root cause

`build_restored_macos_tab` unconditionally used `.incognito(true)`. The source explicitly documented that authenticated servers would re-prompt after restart, so a valid 30-day WebUI auth cookie was discarded whenever the desktop app restored its session.

## Security

- login cookie values are never logged or written into desktop preferences/session JSON
- donor selection requires exact normalized scheme, host, and effective port
- copied cookies are additionally filtered by destination domain, path, and Secure scope
- only the first restored view uses the persistent store; later views remain isolated
- two fail-closed independent review cycles were run; the first found default-profile and cookie-scope blockers, both fixed; the second reported zero logic/security blockers

## Verification

- `cargo fmt --check`
- `cargo clippy -- -D warnings`
- `cargo test` — 27 passed, 1 intentional live-network test ignored
- `python3 scripts/test_macos_ats.py`
- `plutil -lint src-tauri/Info.plist`
- release build completed and local app bundle passed `codesign --verify --deep --strict`
- live macOS acceptance: logged in once, quit PID 2431, relaunched as PID 2558, restored `/session/1a08bea49b3f` returned HTTP 200 with no new `/api/auth/login` POST

## Test plan

- [x] first restored tab uses the persistent cookie store
- [x] later restored tabs remain incognito
- [x] default-profile restore clears stale profile selection
- [x] cookie donors require exact HTTP(S) origin
- [x] copied cookies honor destination scope
- [x] real app restart remains authenticated


cloow:
Verified head: `f63d1b72f60edf810b9fe30e23fa333d227c4d47`

Evidence:
- macOS `cargo fmt --check` passed
- macOS `cargo clippy -- -D warnings` passed
- macOS `cargo test`: 27 passed, 1 intentional live-network test ignored
- ATS regression and `plutil -lint` passed
- release `.app` built and passed deep/strict codesign verification
- two fail-closed reviews completed; first found profile/origin blockers, both fixed; second found zero logic/security blockers
- live acceptance: one successful login, quit PID 2431, relaunch PID 2558, restored session returned HTTP 200 with no new login POST

The upstream Actions run is `action_required` with zero jobs, so CI requires maintainer approval: https://github.com/hermes-webui/hermes-desktop-rust/actions/runs/33225561394

All contributor-side engineering is complete; awaiting maintainer workflow approval/review.

nesquena-hermes:
## Exact-head review at `f63d1b72f` — source direction is sound; profile mutation must fail closed and multi-tab proof is missing

The core design is sensible:

- first restored macOS content view uses the persistent WKWebsiteDataStore;
- later restored views remain incognito;
- donor selection is exact normalized scheme + host + effective port;
- native WKHTTPCookie conversion supplies a concrete domain for host-only server cookies, so the domain/path/Secure filter can match WebUI's `hermes_session` and `hermes_profile` cookies;
- queued main-thread seed tasks are FIFO, so the first persistent donor is prepared before later incognito readers;
- auth-cookie values remain in WebKit memory/storage and are not written into session JSON;
- profile selector is still the only value persisted in the session blob.

The patch remains changes-needed on two deterministic isolation paths plus evidence.

### 1. Saved profile cannot reliably replace the donor profile cookie

The seed loop (`windows.rs:741-767`) copies every destination-scoped donor cookie, including the donor's `hermes_profile`. Native WKHTTPCookie conversion gives that cookie a concrete host domain.

The next Set branch (`:775-780`) writes `session::profile_cookie(value)`, whose unit test explicitly requires `domain() == None` (`session.rs:373-382`). Locked wry 0.55.1 converts None to an empty `NSHTTPCookieDomain`. Cookie identity is name + domain + path, so the empty-domain cookie cannot replace the copied host-domain donor cookie. A restored B-profile tab can retain A's donor selector, emit competing same-name cookies, or hit wry's native cookie-construction `expect` if Darwin rejects the empty domain.

Before seeding, exclude every `hermes_profile` cookie from the donor copy. Then set the saved selector with the exact target host domain (or delete exact-origin selectors first and set one verified replacement). Add a real Mac restore with donor=A, restored=B and prove only one target-scoped profile cookie exists and B is active.

### 2. Cookie-operation failure navigates under the wrong profile

At `windows.rs:738-807`, every failure is logged and then navigation proceeds:

- donor `cookies()` failure;
- inherited `set_cookie()` failure;
- target profile `set_cookie()` failure;
- default-profile `cookies()` failure;
- profile `delete_cookie()` failure.

This is safe-ish for auth inheritance (a failed auth seed can fall back to login), but it is not safe for profile authority. A failed set can open the deep route under default/stale profile; a failed clear can leave the donor/persistent `hermes_profile` selected. The code then navigates anyway, while the PR promises each restored tab reopens on its saved profile.

Make profile settlement fail closed before navigation:

1. seed same-origin auth cookies, excluding `hermes_profile` from donor copy;
2. set the exact saved profile or delete all destination-scoped profile cookies for default;
3. read back and require the exact expected selector/absence;
4. only then navigate the saved deep URL;
5. on failure, do not open that deep route under unknown profile authority—skip restoration for that tab or open a safe root/login surface with an explicit diagnostic.

Bind the operation to the restored window/tab identity so a late cookie callback cannot navigate a closed/replaced view.

### 3. Current tests do not traverse WebKit/Tauri cookie lifecycle

The added Rust tests cover pure booleans/origin/path predicates. They do not call `WebviewWindow.cookies/set_cookie/delete_cookie`, build persistent/incognito data stores, or prove FIFO settlement.

The reported live acceptance proves one restored tab stays authenticated. It does not prove the new multi-tab/multi-window claim.

Required exact-head Mac matrix:

- first restored persistent tab: default profile and named profile;
- second/third incognito tabs: auth inherited, each exact saved profile, no cross-tab profile bleed;
- later native window group inherits from the same exact-origin donor;
- default-profile tab clears a donor's non-default selector;
- different scheme/host/effective port receives zero donor cookies;
- donor read, auth set, profile set, profile read, and profile delete failures—profile failures never navigate the saved deep route;
- app restart plus reopen proves cookies remain out of session JSON and values/logs;
- exact PR `cargo fmt`, clippy, tests, release build, codesign, and live restore evidence.

Also exercise a cold persistent-store read near wry's one-second native cookie-operation timeout; timeout must not silently navigate under unverified profile ownership.

The upstream workflow is currently `action_required` with zero jobs, so there is no repository-owned exact-head CI result yet. Maintainer approval is required before this can be called gate-clean.

`scripts/test_macos_ats.py` is not wired into `.github/workflows/test.yml`, and it checks source plist bytes rather than the built app bundle. Wire it into the Mac gate and inspect the built bundle's merged Info.plist.

### 4. ATS exception is broad—document it honestly

`NSAllowsArbitraryLoadsInWebContent=true` is scoped to embedded WebKit versus native Rust probes, but it permits arbitrary cleartext subresources inside that WebView, not only localhost/Tailscale/LAN/SSH hosts. Top-level navigation remains host-gated by `navigation_allowed`, which limits the impact. Keep this only as an explicit product tradeoff and add an integration oracle for allowed configured HTTP target plus rejected foreign top-level navigation; the current Python test only asserts a plist boolean.

**Verdict: changes needed.** The cookie-persistence model is promising, but profile identity must settle fail-closed and the claimed restored-tab matrix needs exact Mac execution before merge.

PR #89, opened 2026-09-08
fix(windows): avoid WebView polling during voice input

josephcy95:
## Summary

Avoid synchronous child-WebView polling that can freeze the Windows desktop wrapper while voice-dictation software injects text into a focused Hermes WebUI input.

## Changes

- Remove the periodic URL/session capture, cookie polling, and busy-state `eval()` loop.
- Disable synchronous cookie reads used for profile-dot refresh.
- Preserve event-driven tab/window behavior.

## Testing

This branch was tested with Typeless voice input on Windows. The freeze was reproduced before the change and no longer occurred after the change.

The change is intentionally limited to the Tauri desktop wrapper; it does not modify hermes-webui.

### DIFF
--- a/src-tauri/src/main.rs
+++ b/src-tauri/src/main.rs
@@ -395,47 +395,12 @@ fn main() {
                     updater::spawn_check(&update_handle, false);
                 });
             }
-            // Session autosave (issue #18) + per-tab profile-dot refresh (#26):
-            // one 4s timer. `persist` captures windows/tabs so a frame move or
-            // SPA navigation (no structural event) is reflected; the profile
-            // sweep re-reads each strip tab's cookie so a profile switched
-            // *inside* a tab (an SPA re-render) repaints its dot. Both are cheap
-            // and only act on a real change.
-            {
-                let sess_handle = handle.clone();
-                std::thread::spawn(move || loop {
-                    std::thread::sleep(std::time::Duration::from_secs(4));
-                    // Skip this tick's webview work while a native menu modal
-                    // loop is up (the strip's "⋯" popup). Both `persist` (reads
-                    // each tab's URL) and `recapture_profiles` (reads each tab's
-                    // cookie) marshal into the webviews via the runtime
-                    // dispatcher; on Windows that would re-enter the main thread
-                    // from inside the popup's `TrackPopupMenu` modal loop and
-                    // deadlock the UI (#33). The next tick (once the menu closes)
-                    // catches up — both are idempotent and only act on a change.
-                    if sess_handle
-                        .state::<AppState>()
-                        .menu_open
-                        .load(std::sync::atomic::Ordering::SeqCst)
-                    {
-                        continue;
-                    }
-                    session::persist(&sess_handle);
-                    if strip::enabled() {
-                        strip::recapture_profiles(&sess_handle);
-                    }
-                    // Busy re-poll (issue #74): hidden webviews throttle their
-                    // timers, so BUSY_REPORTER's 700ms interval can go quiet in
-                    // a background tab and leave the spinner stale after the
-                    // session finishes. An eval executes regardless of timer
-                    // throttling and re-reads the live S.busy; the reporter's
-                    // debounce means this emits only on an actual change.
-                    windows::eval_all_content(
-                        &sess_handle,
-                        "window.__hermesReportBusy && window.__hermesReportBusy();",
-                    );
-                });
-            }
+            // Windows/WebView2 compatibility: do not periodically poll child
+            // webviews while a focused input may be receiving synthetic text
+            // from voice-dictation tools. Synchronous URL/cookie/eval calls can
+            // re-enter WebView2's UI thread and freeze the host window.
+            // Session persistence remains event-driven (tab/window/navigation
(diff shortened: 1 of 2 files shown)


nesquena-hermes:
## Review at `b898e4f40` — right instinct, wrong blast radius (platform-scope this)

Thanks @josephcy95 — the Windows freeze is real and your diagnosis is correct: a synchronous WebView2 cookie/`eval` read that re-enters the WebView2 UI thread while voice-dictation software is injecting text into a focused input will deadlock the host window. But the current patch fixes a **Windows-only** WebView2 re-entrancy bug by removing the capability on **all three** platforms, and Linux (WebKitGTK) and macOS (WKWebView) don't share that deadlock. I reproduced the regressions by reading the code and confirmed them with our senior-advisor gate (Codex). Verdict: **changes needed** before this can merge.

### Regressions this introduces on Linux/macOS (and a CI break)

1. **Profile-dot capture is now dead on every platform, not just Windows** — `strip.rs:1289`. The unconditional `let _ = (...); return;` sits at the top of `capture_tab_profile()`, and that function is the single chokepoint for **all** callers, including the event-driven ones your PR description says still work: the `route` bridge handler (`recapture_tab_profile`, fired by SSE-driven/SPA navigation — `bridge.rs:689`) and `on_page_load`. So "Profile dots still update on fresh navigation" isn't what the code does — fresh navigation routes straight into the early `return`. Worse, session restore serializes `tab.profile` (the cookie-derived value), while the `profile` bridge event only updates the display-only `dot_profile` — so with capture disabled, a tab whose profile was switched in-page can **reopen under its previous/default profile**. This is a data-correctness regression, not just a cosmetic dot.

2. **Stale busy spinner returns (issue #74) on Linux/macOS** — `main.rs`. Deleting the 4s timer removes the *only* native caller of `window.__hermesReportBusy`. That eval exists precisely because hidden webviews throttle/suspend their own JS timers, so the 700ms `BUSY_REPORTER` goes quiet in a background tab and the spinner stays stuck after the response finishes. There's no event-driven substitute for this — it's a poll by design.

3. **Session autosave loses the no-structural-event cases** — `main.rs`. `session::persist()` does still fire on structural events (window move/resize at `windows.rs:464`, tab open/close/reorder in `strip.rs`), so #18 is the least affected. But the timer covered exactly the cases with *no* structural event — an SPA navigation, a rapid frame change coalesced away by the single-flight guard, structural churn right before quit. Those are now unsaved on all platforms.

4. **CI will fail on this head** — the original `capture_tab_profile` body below the `return;` is now unreachable, which trips the repo's `cargo clippy -- -D warnings` policy (`unreachable_code`). (The PR also currently shows no CI run yet — mergeStateStatus is UNSTABLE with checks pending.)

### The minimal correct fix — scope the workaround to Windows

Keep your Windows behavior exactly as you have it, but stop punishing the platforms that don't deadlock. Concretely:

- **The 4s timer in `main.rs`:** wrap the whole `std::thread::spawn(move || loop { … })` block in `#[cfg(not(target_os = "windows"))]` so Linux/macOS keep session persist + profile sweep + busy re-poll, and Windows gets your no-timer behavior. (The existing `menu_open` skip inside it already handled the `#33` TrackPopupMenu case; that guard stays with the non-Windows build.)
- **`capture_tab_profile` in `strip.rs`:** make the early return `#[cfg(target_os = "windows")]` only — e.g.
  ```rust
  fn capture_tab_profile(app: &AppHandle, window_label: &str, tab_label: &str) {
      #[cfg(target_os = "windows")]
      {
          // WebView2 synchronous cookie reads can re-enter the UI thread and
          // deadlock while voice-dictation software injects text. Skip the read
          // on Windows; the existing menu_open chokepoint guard (#33) covered the
          // popup case, this covers the AX-injection case.
          let _ = (app, window_label, tab_label);
          return;
      }
      #[cfg(not(target_os = "windows"))]
      {
          // …existing body unchanged…
      }
  }
  ```
  That removes the `unreachable_code` clippy failure (nothing is unreachable — one arm is `cfg`'d out), keeps the WebKitGTK/WKWebView cookie read, and preserves profile-restore correctness off Windows.

### The remaining Windows question — a design call, please weigh in

With the workaround correctly scoped, **Windows still loses** session-of-navigation autosave, the busy re-poll, and cookie-based profile capture. If Windows parity matters (I'd argue the profile-restore correctness one does), the right follow-up is a non-synchronous path on Windows — e.g. a debounced persist off cached in-memory state (URL map + strip state you already keep) that never touches `wv.cookies()`/`eval` on the UI thread, and driving the busy indicator from the bridge event rather than an eval poll. That's more than a mechanical change, so I'd rather you own the Windows-parity decision than have me guess it.

One thing I **can't** verify from here: that your scoped Windows change actually clears the Typeless freeze — I don't have a Windows + voice-dictation setup to reproduce it. The `#[cfg]` scoping above preserves your Windows behavior byte-for-byte, so if it fixed the freeze before, it still does; please just re-confirm on your machine after scoping.

Happy to push the mechanical `#[cfg]` scoping to your branch if you'd like — say the word. Otherwise, drop the platform-scoped version here and I'll re-gate the new head.

